PR #1, opened 2020-03-19
[KOGITO-1493] - Dockerfile to build Jenkins Slave Image that supports Golang and Podman

Kaitou786:
https://issues.redhat.com/browse/KOGITO-1493
[KOGITO-1493] - Dockerfile to build Jenkins Slave Image that supports Golang and Podman

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,47 @@
+FROM registry.redhat.io/openshift4/ose-jenkins-agent-base
+
+LABEL MAINTAINER bsig-cloud   <bsig-cloud@redhat.com>
+LABEL Build kogito-cloud-operator using podman
+ENV GOROOT=/usr/local/go \
+    GO_VERSION=1.13  \
+    GOPATH=/home/jenkins/go \
+    GO111MODULE=on  \
+    RELEASE_VERSION=v0.15.1 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+LABEL MAINTAINER bsig-cloud   <bsig-cloud@redhat.com>
+LABEL Build kogito-cloud-operator using podman
+ENV GOROOT=/usr/local/go \
+    GO_VERSION=1.13  \
+    GOPATH=/home/jenkins/go \
+    GO111MODULE=on  \
+    RELEASE_VERSION=v0.15.1 \
+    PATH=/home/jenkins/go/bin:/usr/local/go/bin:/usr/bin/bzr-2.7.0/:$PATH
+RUN  wget https://rpmfind.net/linux/fedora/linux/releases/30/Everything/x86_64/os/Packages/k/kernel-headers-5.0.9-300.fc30.x86_64.rpm && \
```
spolti, line 11: this image is based on rhel 7.7, so, we shouldn't use this approach to install things.

please do not use any third party package from another system.
we should use rhel packages. if that's not possible, then we'll need o do it internally or use a community version of this jenkins image.

spolti (changes requested):

PR #4, opened 2020-05-10
[KOGITO-1884] - Clean up old cucumber test namespaces

msmagnanijr:
See: https://issues.redhat.com/browse/KOGITO-1884

Signed-off-by: Mauricio Magnani <mmagnani@redhat.com>

### DIFF
--- a/tools/cleanup-namespace/cleanup.sh
+++ b/tools/cleanup-namespace/cleanup.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+IN=$(kubectl get namespace -o jsonpath='{range .items[*].metadata}{.name}{";"}{.creationTimestamp}{"\n"}{end}' | awk '/cucumber-/')
+projects=$(echo $IN | tr "\n" "\n")
+
+date_cleanup=$(date -d "1 day ago" +%s)
+
+for values in $projects
+do
+    name_project=${values%;*}
+    date_project=${values#*;}
+
+    if [[ "$date_project" < "$date_cleanup"  ]] ; then
+        oc delete project $name_project
+    fi
+done


tools/cleanup-namespace/cleanup.sh
```diff
@@ -0,0 +1,16 @@
...
+
+date_cleanup=$(date -d "1 day ago" +%s)
+
+for values in $projects
+do
+    name_project=${values%;*}
+    date_project=${values#*;}
+
+    if [[ "$date_project" < "$date_cleanup"  ]] ; then
+        oc delete project $name_project
```
radtriste, line 14: can we add a small `echo` to display which project is deleted ?

radtriste (commented):
Tested briefly and did not work...

here where the displayed entries:
```
date_cleanup = 1589105345
name_project = tradisso-test
date_project = 2020-05-04T08:58:51Z
```

Obviously `date_project` is not in the same format as `date_cleanup`.

```
$ kubectl get namespace tradisso-test -o yaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/description: ""
    openshift.io/display-name: ""
  creationTimestamp: "2020-05-04T08:58:51Z"
  name: tradisso-test
  resourceVersion: "15180669"
  selfLink: /api/v1/namespaces/tradisso-test
  uid: 8a860831-a8a2-4228-be44-9d2a36e937c9
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

tools/cleanup-namespace/cleanup.sh
msmagnanijr: @radtriste sure, i will add the echo. I'm sorry, I should have left this as a draft, I'm still testing it.

spolti: @msmagnanijr let us know when you finish your changes.
you can tag this jira as draft as well.
thanks.

spolti:
@msmagnanijr do you have updates on this?

radtriste:
I guess you could also create a Jenkinsfile that will call this script and that we can then use in a Jenkins job.
wdyt ?

PR #7, opened 2020-05-26
[KOGITO-2293] - Setup job to cleanup old images nightlies

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-2293

Signed-off-by: TARUN KHANDELWAL <tarkhand@redhat.com>

### DIFF
--- a/tools/clean-nightly-tags.py
+++ b/tools/clean-nightly-tags.py
@@ -1,0 +1,89 @@
+#!/usr/bin/python3
+# Script responsible for removing extra tags of nightly images
+# QUAY_ACCESS_TOKEN is needed to set as environment variable before executing script
+# The access token is used for authentication against the quay api.
+
+import os
+import json
+import requests
+from dateutil.relativedelta import *
+from dateutil.easter import *
+from dateutil.rrule import *
+from dateutil.parser import *   
+from datetime import *
+import argparse
+
+
+try:
+    QUAY_ACCESS_TOKEN = os.environ['QUAY_ACCESS_TOKEN']
+except KeyError as e:
+    print("QUAY_ACCESS_TOKEN environment variable is not set. Please, set it before running the script.")
+    exit('Script exiting....')
+
+REGISTRY  = "quay.io"
+NAMESPACE = "kiegroup"
+
+IMAGES={"kogito-data-index-nightly","kogito-quarkus-ubi8-nightly",
+        "kogito-quarkus-jvm-ubi8-nightly","kogito-quarkus-ubi8-s2i-nightly",
+        "kogito-springboot-ubi8-nightly","kogito-springboot-ubi8-s2i-nightly",
+        "kogito-jobs-service-nightly","kogito-management-console-nightly",
+        "kogito-cloud-operator-nightly"
+    }
+def get_image_tags(image):
+    '''
+    Get all the available tags for the image
+    :param image: image name whose tags needs to be fetched
+    :return: tags: List of a strcut with tagName and lastModified as fields
+    '''
+    tags = []
+    r = requests.get('https://{0}/api/v1/repository/{1}/{2}/tag/?onlyActiveTags=true'.format(REGISTRY,NAMESPACE,image) , headers={'content-type': 'application/json', 'Authorization': 'Bearer ' + QUAY_ACCESS_TOKEN })
+    image_metadata= json.loads(r.text)
+    num_tags = len(image_metadata['tags'])
+    for i in range(num_tags):
+        tags.append({
+            "tagName" : image_metadata['tags'][i]['name'], 
+            "lastModified" : parse(image_metadata['tags'][i]['last_modified'])
+        })
+
+    return tags
+
+def delete_image_tags(image, tags):
+    '''
+    Deletes the extra image tags from the repository
+    :param image: Image whose tags needs to be deleted
+    :param tags: List of struct with `tagName` and `last_modified` as fields for the image that needs to be deleted
+    '''
+    if len(tags) == 0:
+        print("Image {} does not have extra tags that needs to be deleted".format(image))
+    else:
+        for tag in tags:
+            requests.delete('https://{0}/api/v1/repository/{1}/{2}/tag/{3}'.format(REGISTRY,NAMESPACE,image,tag['tagName']) , headers={'content-type': 'application/json', 'Authorization': 'Bearer ' + QUAY_ACCESS_TOKEN })
+        print("Successfully deleted {} tags for the image {}".format(len(tags),image))
+
+
+def get_and_delete_old_tags(image,max_tags):
+    '''
+    Driver function, calls the `get_image_tags` to get all the available tags for a image
+    finds the tags that needs to be deleted and then passes them to `delete_image_tags`
+    :param image: image name whose old tags needs to be deleted
+    :param max_tags: Number of maximum tags to be kept for the image
+    '''
+    all_tags = get_image_tags(image)
+    
+    all_tags = list(filter(lambda tag: tag["tagName"]!="latest", all_tags))   #Filter out the entry with latest as tagName from the struct list
+    all_tags.sort(key=lambda tagInfo: tagInfo.get("lastModified"))  #sorting in ascending order to get oldest tag on top
+    delete_tags = []
+    if (len(all_tags) - max_tags) > 0:
+        delete_tags = all_tags[:len(all_tags) - max_tags]
+    
+    delete_image_tags(image,delete_tags)
+    
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Removes extra tags from the registry')
+    parser.add_argument('--max-tags', dest='max_tags', default=10,type=int, help='Defines the maximum number of tags for the image to be available, defaults to 10')
+    args = parser.parse_args()
+
+    for image in IMAGES:
+        get_and_delete_old_tags(image,args.max_tags)


tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,67 @@
...
+
+import os
+import json
+import requests
+import argparse
+
+try:
+    QUAY_TOKEN = os.environ['QUAY_ACCESS_TOKEN']
+except KeyError as e:
+    print("QUAY_ACCESS_TOKEN enviornment variable is not set. Please, set and export on shell and run the script again")
```
radtriste:
```suggestion
    print("QUAY_ACCESS_TOKEN environment variable is not set. Please set it before running the script.")
```

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,67 @@
...
+    exit('Script exciting....')
```
radtriste:
```suggestion
    exit('Script exiting....')
```

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/python3
+# Script responsible for removing extra tags
+# QUAY_ACCESS_TOKEN is needed to set as enviornment variable before executing script
```
radtriste:
```suggestion
# QUAY_ACCESS_TOKEN is needed to set as environment variable before executing script
```

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,67 @@
...
+# The access token is used for authentication against the quay apis.
```
radtriste:
```suggestion
# The access token is used for authentication against the quay api.
```

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,67 @@
...
+        print("Successfully deleted {} tags to keep {} tags for the image {}".format(deleted_tags,max_tags,image))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Removes extra tags from the registry')
+    parser.add_argument('--max-tags', dest='max_tags', default=10,type=int, help='Defines the maximum number of tags for the image to be available, defaults to 10')
+    args = parser.parse_args()
+
+    for image in IMAGES:
+        delete_image_tags(image,args.max_tags)
```
radtriste: Could we have a method `get_and_delete_old_tags` that will call `get_image_tags` and then call `delete_image_tags` with the exact tags to delete ?

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,67 @@
...
+    :param max_tags: Maximum number of tags to keep for the image
+    '''
+    
+    tags = get_image_tags(image)
+    initial_num_tags=len(tags)
+    if initial_num_tags <= max_tags:
+        print("Already {} tags available for image {} which is less or equal to the max_tags: {}".format(initial_num_tags,image,max_tags))
+    else:
+        while len(tags) > max_tags:
+            requests.delete('https://{0}/api/v1/repository/{1}/{2}/tag/{3}'.format(REGISTRY,NAMESPACE,image,tags[-1]) , headers={'content-type': 'application/json', 'Authorization': 'Bearer ' + QUAY_TOKEN })
```
radtriste: Can't we just loop on tags ?

radtriste (changes requested):

tools/clean-nightly-tags.py
Kaitou786: So do you mean passing the tags with the image in `delete_image_tags` as a parameter?

tools/clean-nightly-tags.py
Kaitou786: I think this is better, we are just ensuring here that available tags for the image are not greater than max_tags to keep

tools/clean-nightly-tags.py
radtriste: yes. we determine beforehand which tags to delete and then we make the call

tools/clean-nightly-tags.py
radtriste: that you can do the first time.
I don't get why you need to call back each time to get all tags whereas you already have them

Kaitou786: Just to ensure that are also deleted on quay, but okay, if we assume that they are going to be deleted than yes I can just pass the tags and not get the tags again and again.

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,80 @@
...
+    '''
+    Driver function, calls the `get_image_tags` to get all the available tags for a image
+    passes the tags that needs to be deleted to `delete_image_tags`
+    :param image: image name whose old tags needs to be deleted
+    :param max_tags: Number of maximum tags to be kept for the image
+    '''
+    delete_tags = [] #List of all tags that needs to be deleted
+    
+    all_tags = get_image_tags(image)
+    if len(all_tags) > max_tags:
```
spolti: this will delete any tag depending on the list order quay returns, we need to delete the oldest images and keep only 10.
I think the content returned by quay , besides the tag also contains a field that tells you how old are the tag.

spolti (changes requested):

tools/clean-nightly-tags.py
Kaitou786: Quay returns the list according to the last modification on tag. So the last values will always be the oldest. The same order when we see them on  GUI on tags section.

radtriste: we could, for the safe's side, order them by date somewhere.
If tomorrow Quay changes its API behavior, then we could have some problems ;)

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,80 @@
...
+    Get all the available tags for the image
+    :param image: image name whose tags needs to be fetched
+    :return: tags: List of all the tags available for the image
+    '''
+    tags = []
+    r = requests.get('https://{0}/api/v1/repository/{1}/{2}/tag/?onlyActiveTags=true'.format(REGISTRY,NAMESPACE,image) , headers={'content-type': 'application/json', 'Authorization': 'Bearer ' + QUAY_TOKEN })
+    image_metadata= json.loads(r.text)
+    num_tags = len(image_metadata['tags'])
+    for i in range(num_tags):
+        tags.append(image_metadata['tags'][i]['name'])
```
radtriste: we should get also the last updated date.
Also, we should not count `latest` in here

Kaitou786: Okay I'll create a struct with their date as well,
So apart from `latest` tag we need to save 10 more.

tools/clean-nightly-tags.py
spolti: there is a last modified fied we could use:
`"tags": {"nightly-2020-05-25": {"image_id": "8d034a7aa8c35941a36762e5fb7b7e80b60c2a121c61219d81dd560e1298d8a9", "last_modified": "Mon, 25 May 2020 10:54:12 -0000", "name": "nightly-2020-05-25", "manifest_digest": "sha256:9a80faddd6a391f8fbf105bd4f6abecc52a6d8bf296c98e79876156c3e024498", "size": 105741819}`

Kaitou786: Yeah I used that field only for the timestamp. Can you please recheck the code?

spolti (approved):
sounds better.
thanks

msmagnanijr (approved):

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,101 @@
...
+    Get all the available tags for the image
+    :param image: image name whose tags needs to be fetched
+    :return: tags: Dictionary of all the tags as key and timestame as value available for the image
+    '''
+    tags = {}
+    r = requests.get('https://{0}/api/v1/repository/{1}/{2}/tag/?onlyActiveTags=true'.format(REGISTRY,NAMESPACE,image) , headers={'content-type': 'application/json', 'Authorization': 'Bearer ' + QUAY_TOKEN })
+    image_metadata= json.loads(r.text)
+    num_tags = len(image_metadata['tags'])
+    for i in range(num_tags):
+        tags[image_metadata['tags'][i]['name']] = parse(image_metadata['tags'][i]['last_modified'])
```
radtriste: why not using a struct with `tagName` and `lastUpdate` fields, construct a list of those structs and then use the `sort()` method to order like we want ?
I think the code would be also more readable than just a map

Kaitou786: Personally I don't think that any structure would be better over here than a dictionary with format `tag`: `last_modified`. I did try to use sort function on the value of this dictionary and later also on the dates list (with all the parsed dates) but they didn't result in correct/expected order. The `<` `>` operator was working fine on these values so I used min function which uses `<` to find the minimum value.

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,96 @@
...
+def get_and_delete_old_tags(image,max_tags):
+    '''
+    Driver function, calls the `get_image_tags` to get all the available tags for a image
+    finds the tags that needs to be deleted and then passes them to `delete_image_tags`
+    :param image: image name whose old tags needs to be deleted
+    :param max_tags: Number of maximum tags to be kept for the image
+    '''
+    delete_tags = [] #List of all tags that needs to be deleted
+    all_tags = get_image_tags(image)
+    for i in range(len(all_tags)):
```
radtriste: Please use `filter` method on list: https://www.programiz.com/python-programming/methods/built-in/filter
So it could be later extended if needed

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,96 @@
...
+        if all_tags[i]['tagName'] == 'latest':
+            latest_tag_index = i
+            break
+    all_tags.pop(latest_tag_index)
+
+    all_tags.sort(key=lambda tagInfo: tagInfo.get("lastModified"))  #sorting in ascending order to get oldest tag on top
+    num_delete_tags = len(all_tags) - max_tags #Get the number of tags that needs to be deleted
```
radtriste: you could use `all_tags[:len(all_tags) - max_tags]` to have the subset

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,96 @@
...
+    if num_delete_tags > 0:
+        for i in range(num_delete_tags):
+            delete_tags.append(all_tags[i]['tagName'])
+    
+    delete_image_tags(image,delete_tags)
```
radtriste, line 79: please send the whole list of struct, then the `delete_image_tags` decide which information is useful

radtriste (changes requested):

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,99 @@
...
+        "kogito-jobs-service-nightly","kogito-management-console-nightly",
+        "kogito-cloud-operator-nightly"
+    }
+def filter_latest_tag(all_tags):
+    '''
+    Used for filtering the latest tag from the list 
+    False returns are filtered out by the inbuilt function `filter
+    :param all_tags: list of struct of all tags with `tagName` and `last_modified` as fields
+    '''
+    if all_tags['tagName'] == 'latest':
```
radtriste: can't we just return `!all_tags['tagName'] == 'latest'` ?

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,99 @@
...
+REGISTRY  = "quay.io"
+NAMESPACE = "kiegroup"
+
+IMAGES={"kogito-data-index-nightly","kogito-quarkus-ubi8-nightly",
+        "kogito-quarkus-jvm-ubi8-nightly","kogito-quarkus-ubi8-s2i-nightly",
+        "kogito-springboot-ubi8-nightly","kogito-springboot-ubi8-s2i-nightly",
+        "kogito-jobs-service-nightly","kogito-management-console-nightly",
+        "kogito-cloud-operator-nightly"
+    }
+def filter_latest_tag(all_tags):
```
radtriste: you are filtering on one tag here

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,99 @@
...
+from dateutil.relativedelta import *
+from dateutil.easter import *
+from dateutil.rrule import *
+from dateutil.parser import *   
+from datetime import *
+import argparse
+
+
+try:
+    QUAY_TOKEN = os.environ['QUAY_ACCESS_TOKEN']
```
radtriste: you are checking for `QUAY_ACCESS_TOKEN` but use `QUAY_TOKEN` when making http request ...

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,99 @@
...
+def get_and_delete_old_tags(image,max_tags):
+    '''
+    Driver function, calls the `get_image_tags` to get all the available tags for a image
+    finds the tags that needs to be deleted and then passes them to `delete_image_tags`
+    :param image: image name whose old tags needs to be deleted
+    :param max_tags: Number of maximum tags to be kept for the image
+    '''
+    all_tags = get_image_tags(image)
+    
+    all_tags = list(filter(filter_latest_tag,all_tags))
```
radtriste: can't we just do `list(filter(lambda tag: tag["tagName"]!="latest", all_tags))` ?

radtriste (changes requested):

Kaitou786:
I'll make the changes and test the code, once quay.io is live

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,89 @@
...
+def get_and_delete_old_tags(image,max_tags):
+    '''
+    Driver function, calls the `get_image_tags` to get all the available tags for a image
+    finds the tags that needs to be deleted and then passes them to `delete_image_tags`
+    :param image: image name whose old tags needs to be deleted
+    :param max_tags: Number of maximum tags to be kept for the image
+    '''
+    all_tags = get_image_tags(image)
+    
+    all_tags = list(filter(lambda tag: tag["tagName"]!="latest" ,all_tags))   #Filter out the entry with latest as tagName from the struct list
```
radtriste:
```suggestion
    all_tags = list(filter(lambda tag: tag["tagName"]!="latest", all_tags))   #Filter out the entry with latest as tagName from the struct list
```

radtriste (changes requested):

tools/clean-nightly-tags.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/python3
+# Script responsible for removing extra tags
```
radtriste:
```suggestion
# Script responsible for removing extra tags of nightly images
```

radtriste (approved):
lgtm

spolti:
@Kaitou786 this PR the changes are good, but, shouldn't  the Nightly job be updated as well to call this script?

Kaitou786:
@spolti I was thinking, we can have a periodic Jenkins job which runs this script every morning or after the nightly is complete and delete the images. But it can also be the last step in the nightly pipeline after the images have been published. We need to decide which approach we are going to follow. @radtriste @msmagnanijr  please share your thoughts on this.

spolti:
It can be a something on the nightly pipeline, after push the new images, delete the oldest ones.
we just need to make sure that these changes will not be lost when we migrate it to the rhba jenkins.

radtriste:
IMO, the cleanup is not the responsibility of the Nightly. The cleanup can be done as we want, when we want and where we want.
The Nightly is only responsible for building, testing and deploying.
So I would go like @Kaitou786 suggested in a separate job. We could then customize it like we want it.
But it should anyway NOT be called by the nightly orchestrating pipeline. It should remain independent.

Kaitou786:
So can we merge this one, after that I'll setup a jenkins job to run this script pointing at this repository?

spolti:
ok, I'll merge this one., but keep the job on a place that we'll not loose when migrating the Jenkins.;

PR #9, opened 2020-06-03
KOGITO-2293 - Setup job to cleanup old images nightlies

Kaitou786:
Jenkinsfile for running the script

Signed-off-by: TARUN KHANDELWAL <tarkhand@redhat.com>

### DIFF
--- a/clean-nightly-images-Jenkinsfile
+++ b/clean-nightly-images-Jenkinsfile
@@ -1,0 +1,21 @@
+pipeline{
+    agent {label 'jenkins-slave'}
+    environment {
+        QUAY_ACCESS_TOKEN = credentials("QUAY_ACCESS_TOKEN")
+    }
+    stages{
+        stage('Clone Repo'){
+            steps{
+                sh "rm -rf kogito-pipelines"
+                sh "git clone https://github.com/kiegroup/kogito-pipelines"
+            }
+        }
+        stage('Clean Images'){
+            steps{
+                dir('kogito-pipelines'){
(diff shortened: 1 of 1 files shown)


clean-nightly-images-Jenkinsfile
```diff
@@ -0,0 +1,21 @@
+pipeline{
+    agent {label 'jenkins-slave'}
+    environment {
+        QUAY_ACCESS_TOKEN = credentials("QUAY_ACCESS_TOKEN")
+    }
+    stages{
+        stage('Clone Repo'){
+            steps{
+                sh "rm -rf kogito-pipelines"
+                sh "git clone https://github.com/kiegroup/kogito-pipelines"
```
radtriste: Could you use https://www.jenkins.io/doc/pipeline/steps/workflow-scm-step/ instead ?
Example: https://github.com/kiegroup/kogito-runtimes/blob/fa0257de39b1727d52989ffc06664ee2c87d6fd2/Jenkinsfile.deploy#L42

radtriste (changes requested):

spolti (approved):
approving, but, merge after Tristan suggestion is addressed.

clean-nightly-images-Jenkinsfile
Kaitou786: It turns out no need to checkout the scm, Jenkins will do that for us as the script is in the same repository as pipeline

radtriste (approved):

PR #39, opened 2020-06-19
[KOGITO-2451] Migrate nightly pipelines to RHBA

radtriste:
https://issues.redhat.com/browse/KOGITO-2451

Also 
- https://issues.redhat.com/browse/KOGITO-2569

Need to be merged in sync with https://github.com/kiegroup/kogito-images/pull/182 and https://github.com/kiegroup/kogito-cloud-operator/pull/386

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -1,7 +1,16 @@
-REMOTE_JOB_TIMEOUT=21600 // Timeout to 6 hours (some jobs take time to start)
-WAIT_FAIL_AFTER_X_ERRORS=60
+@Library('jenkins-pipeline-shared-libraries')_
 
-BUILD_URLS=[]
+RUNTIMES_DEPLOY = "kogito-runtimes-deploy"
+IMAGES_DEPLOY = "kogito-images-deploy"
+OPERATOR_DEPLOY = "kogito-operator-deploy"
+RUNTIMES_PROMOTE = "kogito-runtimes-promote"
+IMAGES_PROMOTE = "kogito-images-promote"
+OPERATOR_PROMOTE = "kogito-operator-promote"
+
+// Map of execute jobs
+// See https://javadoc.jenkins.io/plugin/workflow-support/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.html
+// for more options on built job entity
+JOBS = [:]
 
 // Should be multibranch pipeline
 pipeline {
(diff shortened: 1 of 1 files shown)


Jenkinsfile.nightly
```diff
@@ -144,189 +166,68 @@ pipeline {
...
-String constructKey(String prefix, String paramId){
-    if (prefix == null || prefix == ''){
-        return paramId
+boolean isJobSucceeded(String jobName){
+    def job = getJob(jobName)
+    if(job != null){
+        return job.getResult() == "SUCCESS"
     }
-    return "${prefix}_${paramId}"
+    return false
```
Ginxo: It could be simplify by this, just a suggestion
```
def job = getJob(jobName)
return job ? job.getResult() == "SUCCESS" : false
```

Jenkinsfile.nightly
```diff
@@ -144,189 +166,68 @@ pipeline {
...
-            errorCount++
-            if(errorCount >= WAIT_FAIL_AFTER_X_ERRORS){
-                error "Error count has been reached. Failing waitFor loop ..."
-            }
-        }
-
-        sh '#!/bin/sh -e\n' + "sleep ${checkInterval}s"
-        timeout += checkInterval
+String constructKey(String prefix, String paramId){
+    if (prefix == null || prefix == ''){
```
Ginxo: it could be simplify by this (just a suggestion)
` return prefix == null || prefix == '' ? paramId : "${prefix}_${paramId}"`

Ginxo (approved):
apart from the suggestions it LGTM

Jenkinsfile.nightly
```diff
@@ -13,32 +22,33 @@ pipeline {
...
+    }
+
     environment {
         KOGITO_CI_EMAIL_TO = credentials('KOGITO_CI_EMAIL_TO')
 
-        // This will be deleted once we move all jobs to main jenkins
-        JENKINS_CLOUD_ROOT_URL = "https://jenkins-kogito-tools.apps.kogito.automation.rhmw.io/job/Kogito"
-        JENKINS_CLOUD_CREDS = credentials('JENKINS_CLOUD_KOGITO_CREDS')
-        JENKINS_CLOUD_JOB_REMOTE_TOKEN = credentials('JENKINS_CLOUD_KOGITO_JOBS_REMOTE_TOKEN')
+        MAVEN_REPOSITORY = "https://origin-repository.jboss.org/nexus/content/groups/public/"
```
Kaitou786, line 36: Do you thing apart from MAVEN_REPOSITORY we should define a parameter with SNAPSHOT_VERSION(the one which has `8.0.0-SNAPSHOT` as default), that could give us more flexibility or I don't know if it's going to change or have this only for the latest artifacts?

radtriste: What do you mean exactly ? did not get that one ...

Jenkinsfile.nightly
radtriste: true, will update that

Jenkinsfile.nightly
radtriste: right, updating

Jenkinsfile.nightly
Kaitou786: I mean for instance for data-index the path would become https://origin-repository.jboss.org/nexus/content/groups/public/org/kie/kogito/data-index/8.0.0-SNAPSHOT/ but if you see there are other version https://origin-repository.jboss.org/nexus/content/groups/public/org/kie/kogito/data-index/ in here as well. So I want to ask if the `8.0.0-SNAPSHOT`  will always have the latest deployed artifacts, or it can change in the future, We have a option in the script to change this version, do you think we should have this argument in our jenkins nightly setup?

radtriste: oh I see. All repositories should be configured to work with correct version of other repos.
For `8.0.0-SNAPSHOT` is our default versioning for runtimes. So 8.0.0-SNAPSHOT is the default in each repository needing the runtimes.
The update of version will be done in the release pipeline, not in the nightly pipeline.

Kaitou786 (approved):

PR #47, opened 2020-08-10
KOGITO-3028 Rm specific repository for artifacts

radtriste:
With https://issues.redhat.com/browse/KOGITO-3028, artifacts will be sent to origin-repository directly and there should not be any conflicts on `repository.jboss.org` which should be the default used.

So no need to specify the origin repo anymore.

Keeping as draft for now, as I would like to run some tests first in the next days


Ginxo (approved):
LGTM

Kaitou786 (approved):

PR #59, opened 2020-09-14
Missing default build parameters for create release branch

cristianonicolai:


evacchi (approved):

radtriste:
@cristianonicolai This had an impact on the release ?

radtriste (approved):

radtriste:
In fact `DISPLAY_NAME` and `RELEASE_BRANCH` are already set and others have default values... What was the failure there ?

cristianonicolai:
@radtriste we were getting:
`NoSuchMethodError: No such DSL method 'addStringParam' found among steps`

Jenkinsfile.release
```diff
@@ -94,7 +94,7 @@ pipeline {
                 script {
                     echo "Call ${CREATE_RELEASE_BRANCHES} job"
 
-                    def buildParams = [:]
+                    def buildParams = getDefaultBuildParams()
```
radtriste, line 97: Oh I see now ...
Correction should just have been this one. No need to use `getDefaultBuildParams` and put all other parameters in the call.

```suggestion
                    def buildParams = []
```

@cristianonicolai Could be set like that after the release IMO ;)

PR #85, opened 2020-11-09
KOGITO-3800 Provide mapping between kogito and optaplanner

rsynek:
https://issues.redhat.com/browse/KOGITO-3800

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -102,7 +102,20 @@ pipeline {
             steps {
                 script {
                     echo "Call ${OPTAPLANNER_DEPLOY} job"
-                    def buildParams = getDefaultBuildParams()
+
+                    String optaplannerBranch = ''
+                    if (getBuildBranch() != 'master') {
+
+                        /* The OptaPlanner major version is shifted by 7 from the Kogito major version:
+                           Kogito 1.x.y -> OptaPlanner 8.x.y. */
+                        int majorVersionShift = 7
+                        String [] buildBranchSplit = getBuildBranch().split("\\.")
+                        assert buildBranchSplit.length == 3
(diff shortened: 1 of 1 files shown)


Kevin-Mok (approved):

Jenkinsfile.nightly
```diff
@@ -107,6 +107,19 @@ pipeline {
...
+
+                        /* The OptaPlanner major version is shifted by 7 from the Kogito major version:
+                           Kogito 1.x.y -> OptaPlanner 8.x.y. */
+                        int majorVersionShift = 7
+                        String [] buildBranchSplit = getBuildBranch().split("\\.")
+                        assert buildBranchSplit.length == 3
+                        Integer optaplannerMajorVersion = Integer.parseInt(buildBranchSplit[0]) + majorVersionShift
+                        String optaplannerBranch = "${optaplannerMajorVersion}.${buildBranchSplit[1]}.${buildBranchSplit[2]}"
+
+                        addStringParam(buildParams, 'BUILD_BRANCH_NAME', optaplannerBranch)
```
radtriste: This is in fact not replacing the other `BUILD_BRANCH_NAME` param set in `getDefaultBuildParams()` method but just adding it.
This could lead to some problems when launching the job (even if for now, the latest param found is taken into account).

Could we leverage something similar to what has been updated on release pipeline ? like  https://github.com/kiegroup/kogito-pipelines/blob/master/Jenkinsfile.release#L562 without the version param of course.
or just in that job copy the content of `getDefaultBuildParams()` and do not call it ?

radtriste (approved):

Jenkinsfile.nightly
rsynek: Fixing as you suggest for now; later we might wanna look into https://issues.redhat.com/browse/KOGITO-3843.

rsynek:
Running a final test.

PR #96, opened 2020-12-14
Create new release branch for optaplanner quickstarts based upon development

dupliaka:
OptaPlanner quick starts use development instead of master as a base for the next release branch. 
To be able to customize that I add one more parameter NON_MASTER_BASES.
User can specify for different repositories different bases in format repo1:branch1,repo2:branch2
By default if the parameter was not passed or it is empty - then master will be used

The test job could be found here https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/adupliak/job/quickstarts-release-pipeline/job/kogito-release/11/

### DIFF
--- a/Jenkinsfile.create-release-branches
+++ b/Jenkinsfile.create-release-branches
@@ -1,5 +1,7 @@
 @Library('jenkins-pipeline-shared-libraries')_
 
+CUSTOM_BASES=[:]
+
 pipeline {
     agent {
         label 'kie-rhel7 && !master'
@@ -11,6 +13,11 @@ pipeline {
         
         string(name: 'GIT_AUTHOR', defaultValue: 'kiegroup', description: 'Which Git author repository ?')
         string(name: 'GIT_AUTHOR_CREDS_ID', defaultValue: 'kie-ci', description: 'Git author credentials id')
+
+        string(name: 'NON_MASTER_BASES', defaultValue: '', description: 'Comma-separated list of custom bases in format repo:base')
+    }
+    environment {
+        DEFAULT_BASE_BRANCH = 'master'
     }
 
     stages {
@@ -23,6 +30,7 @@ pipeline {
 
                     currentBuild.displayName = "${getReleaseBranch()}"
                     currentBuild.description = "${getRepositories()}"
+                    CUSTOM_BASES = getCustomBasesMap()
                 }
             }
         }
@@ -36,7 +44,7 @@ pipeline {
                         echo "Checkout repo ${repo}"
                         dir(repo){
                             deleteDir()
-                            checkout(githubscm.resolveRepository(repo, getGitAuthor(), 'master', false))
+                            checkout(githubscm.resolveRepository(repo, getGitAuthor(), getBaseBranch(repo), false))
                             sh 'git fetch origin'
                             String branchRemoteResult = sh(script: "git ls-remote origin ${getReleaseBranch()} | wc -l", returnStdout: true).trim()
                             if(Integer.parseInt(branchRemoteResult) > 0){
@@ -73,4 +81,17 @@ String getGitAuthorCredsId(){
 
 String[] getRepositories(){
     return params.REPOSITORIES.split(',')
+}
+
+String getBaseBranch(String repo) {
+    return (CUSTOM_BASES && CUSTOM_BASES.containsKey(repo))?CUSTOM_BASES.get(repo):env.DEFAULT_BASE_BRANCH;
+}
+
+def getCustomBasesMap() {
+    return params.NON_MASTER_BASES
+            .split(', ')
+            .collectEntries { entry ->
+                def pair = entry.split(':')
+                [(pair.first()): pair.last()]
+            }
 }
--- a/Jenkinsfile.create-release-branches
+++ b/Jenkinsfile.create-release-branches
@@ -6,7 +6,7 @@ pipeline {
     }
 
     parameters {
-        string(name: 'REPOSITORIES', defaultValue: '', description: 'Comma-separated list of repositories to update. Add branch name after semicolon to customize base')
+        string(name: 'REPOSITORIES', defaultValue: '', description: 'Comma-separated list of repository[:base branch] to update. The default base branch for every repository is the \'master\' branch.
         string(name: 'RELEASE_BRANCH', defaultValue: '', description: 'Release branch to create')
         
         string(name: 'GIT_AUTHOR', defaultValue: 'kiegroup', description: 'Which Git author repository ?')
--- a/Jenkinsfile.create-release-branches
+++ b/Jenkinsfile.create-release-branches
@@ -1,7 +1,5 @@
 @Library('jenkins-pipeline-shared-libraries')_
 
-CUSTOM_BASES=[:]
-
 pipeline {
     agent {
         label 'kie-rhel7 && !master'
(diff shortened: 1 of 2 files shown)


Jenkinsfile.create-release-branches
```diff
@@ -11,6 +13,11 @@ pipeline {
         
         string(name: 'GIT_AUTHOR', defaultValue: 'kiegroup', description: 'Which Git author repository ?')
         string(name: 'GIT_AUTHOR_CREDS_ID', defaultValue: 'kie-ci', description: 'Git author credentials id')
+
+        string(name: 'NON_MASTER_BASES', defaultValue: '', description: 'Comma-separated list of custom bases in format repo:base')
```
rsynek: The new parameter duplicates the existing `REPOSITORIES` parameter - again, we provide a list of repositories, this time also with the base branch.

Similarly to the base branch, also the release branch name differs across all repositories participating in the release pipeline. To solve this, the `create-release-branches` job is triggered once for every group of repositories with the same release branch name.

We can do the same for the base branch, which would be passed to the job as a simple `String` parameter. Another option is to keep the `repo:branch` format, but directly in the `REPOSITORIES` parameter.

dupliaka: 1.Pass simple string parameter BASE_BRANCH
2.Pass within the repositories

1.Release branch differs not for all repositories but for kogito like 1.0.x and optaplanner like 8.0.x. SO in case if I do it as a parameter I would need to either create a separate step or another buildParam variable to pass the params. Duplication in stage or duplication in buildParam
2.Almost every repo got base master so the list of repos will be: ['kogito-runtimes':'master','kogito-apps':'master','kogito-examples':'master'] and we are not allowed to declare a key without a value.

Jenkinsfile.create-release-branches
```diff
@@ -1,5 +1,7 @@
 @Library('jenkins-pipeline-shared-libraries')_
 
+CUSTOM_BASES=[:]
```
radtriste: `CUSTOM_BASES` is used only in `getBaseBranch` method so I would just use a local var in there.

Jenkinsfile.create-release-branches
radtriste: why not then merging `REPOSITORIES` and `NON_MASTER_BASES` string parameters into a [text parameter](https://www.jenkins.io/doc/book/pipeline/syntax/#parameters) which contains list of properties like:
```
optaplanner=
optaweb-employee-rostering=
optaplanner-quickstarts=development
```
use then [`readProperties`](https://www.jenkins.io/doc/pipeline/steps/pipeline-utility-steps/#readproperties-read-properties-from-files-in-the-workspace-or-text) to get the key/values
and if no branch specified (we can put empty values), then use `master`

Jenkinsfile.release
```diff
@@ -61,13 +62,13 @@ pipeline {
...
         KOGITO_CI_EMAIL_TO = credentials('KOGITO_CI_EMAIL_TO')
-
         // RELEASE_BRANCH => will be set in Initialization phase
 
         // We use quay as temporary registry for testing between the jobs instead of Openshift, due to https://issues.redhat.com/browse/KOGITO-2219
 
-        IMAGE_REGISTRY_CREDENTIALS = 'release_kogito'
-        IMAGE_REGISTRY = 'quay.io'
+        IMAGE_REGISTRY_CREDENTIALS = 'EMPTY_REGISTRY_CREDENTIALS'
+        IMAGE_REGISTRY = 'EMPTY_REGISTRY'
```
radtriste: testing purpose ?

Jenkinsfile.release
```diff
@@ -15,9 +15,10 @@ OPERATOR_PROMOTE = 'kogito-operator-promote'
 
 PIPELINE_REPOS=['kogito-pipelines']
 RUNTIMES_REPOS=['kogito-runtimes','kogito-apps','kogito-examples']
-OPTAPLANNER_REPOS =['optaplanner', 'optaweb-vehicle-routing', 'optaweb-employee-rostering']
+OPTAPLANNER_REPOS =['optaplanner', 'optaweb-vehicle-routing', 'optaweb-employee-rostering', 'optaplanner-quickstarts']
 IMAGES_REPOS=['kogito-images']
 OPERATOR_REPOS=['kogito-cloud-operator']
+NON_MASTER_BASES=['optaplanner-quickstarts':'development']
```
radtriste: I would rename to more meaningful like `REPOSITORIES_BASE_BRANCH`

Jenkinsfile.create-release-branches
rsynek: As for the second option, I think that the `master` branch can be considered a default value, so it's more like follows:
"kogito-runtimes,kogito-apps,kogito-examples" and "optaplanner,optaweb-employee-rostering,optaplanner-quickstarts:development".
@radtriste is proposing a properties format for this parameter, but the principle is the same. The goal is to avoid sending a list of repositories in two parameters.

dupliaka: Ok, no problems. I like @rsynek proposal. Indeed the format is not so important but I will keep it as close to the native dictionaries as possible.

Jenkinsfile.release
dupliaka: Thanks a lot, great catch fixed in b29981a

Jenkinsfile.release
dupliaka: Removed completely in b0d92b2

Jenkinsfile.create-release-branches
dupliaka: Removed  in b0d92b2

Jenkinsfile.create-release-branches
dupliaka: Fixed in  b0d92b2 
Just parse url and branch. If branch is defined use it.

Jenkinsfile.create-release-branches
radtriste: can you please also delete this line ? :)

Jenkinsfile.release
```diff
@@ -713,4 +714,7 @@ String getKogitoOperatorTempTag(){
 String removeVersionSuffixIfExist(String version){
     Integer[] versionSplit = util.parseVersion(version)
     return "${versionSplit[0]}.${versionSplit[1]}.${versionSplit[2]}"
+}
+String getCustomBases(){
```
radtriste: You never call it in the end ?

dupliaka:
Result of b29981a
https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/adupliak/job/quickstarts-release-pipeline/job/kogito-release/14/

radtriste (approved):

Jenkinsfile.create-release-branches
```diff
@@ -6,11 +6,15 @@ pipeline {
     }
 
     parameters {
-        string(name: 'REPOSITORIES', defaultValue: '', description: 'Comma-separated list of repositories to update')
+        string(name: 'REPOSITORIES', defaultValue: '', description: 'Comma-separated list of repositories to update. Add branch name after semicolon to customize base')
```
rsynek:
```suggestion
        string(name: 'REPOSITORIES', defaultValue: '', description: 'Comma-separated list of repository[:base branch] to update. The default base branch for every repository is the \'master\' branch.
```

rsynek: => It's not a semicolon.

rsynek (approved):

PR #115, opened 2021-02-12
Add possibility to generate jobs specific repositories

radtriste:
Useful for testing only specific pipeline

### DIFF
--- a/docs/jenkins.md
+++ b/docs/jenkins.md
@@ -100,7 +100,7 @@ echo '----- Launching seed tests'
 ${TEMP_DIR}/dsl/seed/scripts/seed_test.sh ${TEMP_DIR}
 ```
 
-The script clones the `kogito-pipelines` repository and then call the `seed_test.sh` which should copy the jobs and 
+The script clones the `kogito-pipelines` repository and then call the `seed_test.sh` which should copy the jobs and run the tests.
 
 Then you can call the script:
 
@@ -109,7 +109,7 @@ $ chmod u+x .jenkins/dsl/test.sh
 $ cd .jenkins/dsl && ./scripts/test.sh
 ```
 
-*NOTE: Your Jenkinsfiles should be in `.jenkins` folder but can be anywhere. The reference is anyway done in the job script, so you can put whatever you want if needed.*
+*NOTE: Your Jenkinsfiles can be stored anywhere in the repository. The reference is anyway done in the job script, so you can put whatever you want if needed. One good practise would be to store it at the root of the project or in the `.jenkins` folder.*
 
 ## Test specific jobs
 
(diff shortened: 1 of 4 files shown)


radtriste:
tested all ok and using it :)

docs/jenkins.md
```diff
@@ -114,4 +109,34 @@ $ chmod u+x .jenkins/dsl/test.sh
 $ cd .jenkins/dsl && ./scripts/test.sh
 ```
 
-*NOTE: Your Jenkinsfiles should be in `.jenkins` folder but can be anywhere. The reference is anyway done in the job script, so you can put whatever you want if needed.*
\ No newline at end of file
+*NOTE: Your Jenkinsfiles should be in `.jenkins` folder but can be anywhere. The reference is anyway done in the job script, so you can put whatever you want if needed.*
```
Kaitou786: This Note seems confusing to me, I mean first it is written that `it should be in...` and then can be anywhere. can we rephrase this statement?

dsl/seed/Jenkinsfile.seed
```diff
@@ -45,7 +45,21 @@ pipeline {
...
+                    if (getCustomBranchKey()) {
+                        currentBuild.displayName = getCustomBranchKey()
+
+                        assert getCustomRepositories()
+                        branchConfigs = [ "${getCustomBranchKey()}" : [:] ]
+                        if (getCustomMainBranch()) {
+                            seedConfig.git.main_branch = [ "default" : getCustomMainBranch() ]
+                        }
+                    } else {
+                        branchConfigs = seedConfig.git.branches.collectEntries { [it, [:]] }
```
Kaitou786, line 58: Just out of curiosity what does this `it` object represent here?

Kaitou786 (commented):
Also why do we have these lines commented here: https://github.com/kiegroup/kogito-pipelines/pull/115/files#diff-7fc4c71a02d66fa47c3dece6c2ae2e8745e18501011f2bbf08b4b45f3c5ba0e5R24-R26

dsl/seed/Jenkinsfile.seed
radtriste: `it` represent the iterative object.
you can do many ways in groovy. this way or:
```
seedConfig.git.branches.collectEntries { branch -> [branch, [:]] }
```
one other thing is about iteration on maps:
```
map.each { println "${it.key} = ${it.value}" }

map.each { entry -> println "${entry.key} = ${entry.value}" }

map.each { key, value -> println "${key} = ${value}" }
```
all are valid :D

docs/jenkins.md
radtriste: updated

Kaitou786 (approved):

msmagnanijr (approved):
@radtriste nice work :)

PR #135, opened 2021-03-16
Add SKIP_INTEGRATION_TESTS parameter

dupliaka:
SKIP_INTEGRATION_TESTS would derive value from SKIP_TEST so having SKIP_TEST set to true will skip all tests.

So you run all tests or not running it at all.

@radtriste  WDYT?


radtriste (approved):
I am good with that. For now we have no distinction in nightly for tests and integration tests so seems a good solution

dupliaka:
@radtriste could you please merge this PR?

radtriste (approved):

PR #164, opened 2021-05-10
Correct GHAction

radtriste:

PR #165, opened 2021-05-12
[PR Jobs] non-forked PRs error fixed

Ginxo:


dsl/seed/src/main/groovy/org/kie/jenkins/jobdsl/templates/KogitoJobTemplate.groovy
```diff
@@ -74,10 +74,11 @@ class KogitoJobTemplate {
                     scm {
                         git {
                             remote {
-                                url(jobParams.git.repo_url ?: Utils.createRepositoryUrl('${ghprbPullAuthorLogin}', jobParams.git.repository))
+                                url(jobParams.git.repo_url ?: Utils.createRepositoryUrl('kiegroup', jobParams.git.repository))
```
radtriste: I would use `jobParams.git.author` here instead of harcoded `kiegroup`
we use that flexibility to be able to generate test jobs on different base authors

radtriste (changes requested):
one change otherwise good to me

dsl/seed/src/main/groovy/org/kie/jenkins/jobdsl/templates/KogitoJobTemplate.groovy
Ginxo: done!

radtriste (approved):

radtriste:
Error is unrelated. trying to solve it in this PR https://github.com/kiegroup/kogito-pipelines/pull/164

radtriste (approved):

PR #192, opened 2021-06-15
Secure the release props if unexpected abortion

radtriste:
In case Jenkins has an unexpected problem, it could be that the file is not archived because the main `post` is not executed ...
With that change, each time a build is done, it will archive the release properties file so that we can restart from latest job built if needed.


Ginxo (approved):
LGTM

msmagnanijr (approved):

PR #282, opened 2021-10-08
KOGITO-5937 Correct sonarcloud

radtriste:


radtriste:
merged as else it would block apps PRs ...

PR #306, opened 2021-10-27
Updated timeout value for native checks

radtriste:
Because native is quite expensive... Some repos need more than 360 minutes


Ginxo (commented):
LGTM

PR #312, opened 2021-11-02
Temporary remove docker prune to show docker info

cristianonicolai:


radtriste:
@cristianonicolai We should not remove that prune as it could have some potential problems on GHA. Can you point to failing PR about the initial problem ?

cristianonicolai:
@radtriste lets drop this for now

PR #326, opened 2021-11-16
BXMSPROD-1542 definition file eval expression

Ginxo:
https://github.com/kiegroup/optaweb-vehicle-routing/pull/602

### DIFF
--- a/.ci/jenkins/Jenkinsfile.buildchain
+++ b/.ci/jenkins/Jenkinsfile.buildchain
@@ -78,7 +78,7 @@ pipeline {
             steps {
                 script {
                     if (isSonarCloudAnalysis()) {
-                        def project = (env.BUILDCHAIN_PROJECT ? util.getProjectGroupName(env.BUILDCHAIN_PROJECT) : util.getProjectTriggeringJob())[1]
+                        def project = getProject()
                         // Project should define a `clone` for the current command so that the pom can be found for sonarcloud analysis
                         // Else this next step will fail
                         dir("bc/kiegroup_${project.replaceAll('-', '_')}/${project}") {
@@ -143,6 +143,10 @@ String getMavenTool() {
     return env.BUILD_MAVEN_TOOL?.trim() ?: 'kie-maven-3.8.1'
 }
 
+String getProject() {
+    return (env.BUILDCHAIN_PROJECT ? util.getProjectGroupName(env.BUILDCHAIN_PROJECT) : util.getProjectTriggeringJob())[1]
+}
+
 String getPrType() {
     return env.BUILDCHAIN_PR_TYPE?.trim() ?: 'pr'
 }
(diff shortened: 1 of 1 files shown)


.ci/jenkins/Jenkinsfile.buildchain
```diff
@@ -173,7 +177,7 @@ def getBuildChainAdditionalArguments() {
 }
 
 String getBuildChainConfigBranch() {
-    return env.BUILDCHAIN_CONFIG_BRANCH ?: '\${BRANCH:main}'
+    return env.BUILDCHAIN_CONFIG_BRANCH ?: ['optaplanner', 'optaweb-employee-rostering', 'optaweb-vehicle-routing', 'drools'].contains(getProject()) ? '%{process.env.GITHUB_BASE_REF.replace(/(\d*)\.(.*)\.(.*)/g, (m, n1, n2, n3) => `\${+n1-7}.\${n2}.\${n3}`)}' : '\${BRANCH:main}'
```
radtriste: I would put this list of projects in a variable else it makes it difficult to read ;)

radtriste (changes requested):

.ci/jenkins/Jenkinsfile.buildchain
Ginxo: @radtriste you are right. Thanks. Done! do you think `NOT_KOGITO_PROJECTS` is a proper variable name? Would you propose a better name for this? :thinking:

radtriste: good for me

.ci/jenkins/Jenkinsfile.buildchain
```diff
@@ -173,7 +179,7 @@ def getBuildChainAdditionalArguments() {
 }
 
 String getBuildChainConfigBranch() {
-    return env.BUILDCHAIN_CONFIG_BRANCH ?: '\${BRANCH:main}'
+    return env.BUILDCHAIN_CONFIG_BRANCH ?: NOT_KOGITO_PROJECTS.contains(getProject()) ? '%{process.env.GITHUB_BASE_REF.replace(/(\d*)\.(.*)\.(.*)/g, (m, n1, n2, n3) => `\${+n1-7}.\${n2}.\${n3}`)}' : '\${BRANCH:main}'
```
radtriste, line 182: just thinking again about this change and I think we should not have project specific context in this file ... which means if we want a Jenkins job to work with a mapping, it will need to provide `BUILDCHAIN_CONFIG_BRANCH` in the env of the job (which is in fact already the case).
Else it will be hard to maintain. I would rather have this in DSL than in the runtime.

.ci/jenkins/Jenkinsfile.buildchain
dupliaka: Other ideas: NON_KOGITO_PROJECTS, NON_KOGITOS

Ginxo:
closed by

- https://github.com/kiegroup/drools/pull/4026
- https://github.com/kiegroup/optaplanner/pull/1678
- https://github.com/kiegroup/optaplanner-quickstarts/pull/230
- https://github.com/kiegroup/optaweb-employee-rostering/pull/722
- https://github.com/kiegroup/optaweb-vehicle-routing/pull/607
- https://github.com/kiegroup/kogito-runtimes/pull/1724

PR #329, opened 2021-11-26
build chain updated to v2.6.21

Ginxo:
* additional-flags
* exports execution summary to spreadsheet


radtriste (approved):

jomarko (approved):
I am not expert in the area. I think I aj not proper person for review. Howevr the code changes looks good to me. Especially there is an "example value"  for "additional-flags"

PR #351, opened 2022-01-18
PLANNER-2641 OptaPlanner depends on Drools, not Kogito

triceo:
@radtriste As we discussed. Please double-check that this is correct.

Downstream PRs: 

* https://github.com/kiegroup/kogito-runtimes/pull/1929
* https://github.com/kiegroup/optaplanner/pull/1800

The changes have been tested against the downstream repos and all PR checks were green.
I checked the execution plans, and they did what they were supposed to.

### DIFF
--- a/.ci/jenkins/Jenkinsfile.prod.nightly
+++ b/.ci/jenkins/Jenkinsfile.prod.nightly
@@ -50,7 +50,7 @@ pipeline {
                         pipelineHelper.retry(
                         {
                             def SETTINGS_XML_ID = RHBA_RELEASE_VERSION ? 'kogito-nightly-version' : 'kogito-nightly-main'
-                            def projectCollection = ['drools', 'kogito-runtimes', 'optaplanner', 'kogito-examples', 'optaplanner-quickstarts', 'optaweb-vehicle-routing', 'optaweb-employee-rostering', 'jboss-integration/kogito-rhba']
+                            def projectCollection = ['drools', 'optaplanner', 'optaplanner-quickstarts', 'optaweb-vehicle-routing', 'optaweb-employee-rostering', 'kogito-runtimes', 'kogito-examples', 'jboss-integration/kogito-rhba']
                             def projectVariableMap = [:]
                             def additionalVariables = [
                                 'droolsProductVersion': env.DROOLS_PRODUCT_VERSION, 
--- a/.ci/jenkins/Jenkinsfile.tools.update-quarkus
+++ b/.ci/jenkins/Jenkinsfile.tools.update-quarkus
@@ -29,7 +29,7 @@ pipeline {
         stage('Call update quarkus jobs') {
             steps {
                 script {
-                    [ 'kogito-runtimes', 'kogito-examples', 'optaplanner', 'optaplanner-quickstarts' ].each { repo ->
+                    [ 'optaplanner', 'optaplanner-quickstarts', 'kogito-runtimes', 'kogito-examples' ].each { repo ->
                         launchUpdateQuarkusJob(repo)
                     }
                 }
--- a/.ci/project-dependencies-quarkus.yaml
+++ b/.ci/project-dependencies-quarkus.yaml
@@ -24,6 +24,25 @@ dependencies:
         - kiegroup/optaweb-vehicle-routing
         - kiegroup/optaplanner-quickstarts
 
+  - project: kiegroup/optaplanner
+    dependencies:
+      - project: kiegroup/drools
+    mapping:
+      dependencies:
+        default:
+          - source: (\d*)\.(.*)\.(.*)
+            targetExpression: "process.env.GITHUB_BASE_REF.replace(/(\\d*)\\.(.*)\\.(.*)/g, (m, n1, n2, n3) => `${+n1-7}.${n2}.${n3}`)"
+      dependant:
+        default:
+          - source: (\d*)\.(.*)\.(.*)
+            targetExpression: "process.env.GITHUB_BASE_REF.replace(/(\\d*)\\.(.*)\\.(.*)/g, (m, n1, n2, n3) => `${+n1+7}.${n2}.${n3}`)"
+      exclude:
+        - quarkusio/quarkus
+        - kiegroup/drools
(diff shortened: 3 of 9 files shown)


.ci/jenkins/Jenkinsfile.prod.nightly
```diff
@@ -50,7 +50,7 @@ pipeline {
                         pipelineHelper.retry(
                         {
                             def SETTINGS_XML_ID = RHBA_RELEASE_VERSION ? 'kogito-nightly-version' : 'kogito-nightly-main'
-                            def projectCollection = ['drools', 'kogito-runtimes', 'optaplanner', 'kogito-examples', 'optaplanner-quickstarts', 'optaweb-vehicle-routing', 'optaweb-employee-rostering', 'jboss-integration/kogito-rhba']
+                            def projectCollection = ['drools', 'optaplanner', 'optaplanner-quickstarts', 'optaweb-vehicle-routing', 'optaweb-employee-rostering', 'kogito-runtimes', 'kogito-examples', 'jboss-integration/kogito-rhba']
```
triceo, line 53: I feel that this is missing `kogito-apps`... ?

.ci/jenkins/Jenkinsfile.tools.update-quarkus
```diff
@@ -29,7 +29,7 @@ pipeline {
         stage('Call update quarkus jobs') {
             steps {
                 script {
-                    [ 'kogito-runtimes', 'kogito-examples', 'optaplanner', 'optaplanner-quickstarts' ].each { repo ->
+                    [ 'optaplanner', 'optaplanner-quickstarts', 'kogito-runtimes', 'kogito-examples' ].each { repo ->
```
triceo, line 32: Dtto., isn't this missing `kogito-apps`?

radtriste: kogito-apps does not depend directly on quarkus. the update will be done via kogito-runtimes.

.ci/jenkins/Jenkinsfile.prod.nightly
radtriste: kogito-apps are not yet productized.
Modifying the prod part may require some update in the build-configuration: https://gitlab.cee.redhat.com/middleware/build-configurations/-/tree/master/kogito/nightly
@Ginxo can help on that

.ci/project-dependencies-quarkus.yaml
```diff
@@ -24,43 +24,43 @@ dependencies:
         - kiegroup/optaweb-vehicle-routing
         - kiegroup/optaplanner-quickstarts
 
-  - project: kiegroup/kogito-runtimes
+  - project: kiegroup/optaplanner
     dependencies:
-      - project: quarkusio/quarkus
       - project: kiegroup/drools
```
radtriste, line 30: so optaplanner does not depend on Quarkus anymore ?

.ci/jenkins/Jenkinsfile.prod.nightly
Ginxo: if this is a matter of changing the order it does not have side effects on `build-configurations`. If it's about adding `kogito-apps` to the list the `build-configurations` is already there https://gitlab.cee.redhat.com/middleware/build-configurations/-/blob/master/kogito/nightly/build-config.yaml#L142 (we will have to check, depending on the order, whether the already existing configuration for kogito-apps still valid)

.ci/project-dependencies-quarkus.yaml
Ginxo: +1

triceo: Most likely an oversight, will fix.

.ci/project-dependencies-quarkus.yaml
```diff
@@ -38,7 +39,6 @@ dependencies:
             targetExpression: "process.env.GITHUB_BASE_REF.replace(/(\\d*)\\.(.*)\\.(.*)/g, (m, n1, n2, n3) => `${+n1+7}.${n2}.${n3}`)"
       exclude:
         - quarkusio/quarkus
-        - kiegroup/drools
```
Ginxo: and the mapping is not excluded for drools? :thinking: then whenever we open a optaplanner for 8.x.y branch drools:1.x.y will be taken? :thinking: (and the opposite).
shouldn't it be 8.x.y to 8.x.y and vice versa? @triceo

triceo: Clearly, I do not understand this part well enough. Would you mind explaining?

Ginxo: sure, you have it documented here https://github.com/kiegroup/build-chain-configuration-reader#mapping
So for instance a mapping like this (just an example)
```
- project: kiegroup/optaplanner
  dependencies:
    - project: X    
    - project: Y
  mapping:
    dependencies:
      default:
        - source: 7.x
          target: main
    dependant:
      default:
        - source: main
          target: 7.x
```

- Let's suppose we have a optaplanner PR against `7.x`, this mapping will take `main` for the rest of the projects (thanks to `mapping.dependencies.default`).
- Let's suppose the opposite, a project X against `main` branch, optaplanner will be mapped to `7.x` (thanks to `mapping.dependant.default`).
- Let's suppose now a optaplanner PR against `8.x`, nothing will happen (from mapping POV) because there isn't any coincidence with `mapping.dependencies` for 8.x

Just keep in mind `source` could be RegEx, so you can defined something like
```
  mapping:
    dependencies:
      default:
        - source: .*
          target: main
```
then all branches from optaplanner will be mapped to main for the rest of the projects.

The `mapping.exclude` section is for excluding a project from this mapping mechanism, so in case we add project X to the exclusion, the mapping won't be executed.

Sorry for the long reply, but it is not simple stuff to explain/understand. Just let me know in case you need more clarification.

radtriste (approved):
apart from comment from Enrique, ltgm

triceo:
@Ginxo I have restored the original excludes for each project, which I believe should do the trick. Please take a look again.

radtriste (approved):

PR #352, opened 2022-01-19
Correct release pipeline with job decision persistence

radtriste:
We should remove the job decision if the job gets (re)executed


mbiarnes (approved):
LGTM

PR #353, opened 2022-01-20
Update quarkus tools scripts to be independent from git directly

radtriste:
To be able to reuse the script anywhere, without the Git commands.

New command if you run from scratch would be
```
sh run-git-script-modification -f evacchi -- ./update-quarkus-platform.sh -v 15.0.Final stage
```
Else if you start from an existing repo, it will simply be
```
sh update-quarkus-platform.sh -v 15.0.Final stage
```


radtriste:
Tested for quarkus platform in the latest 2 PRs.
I need to run some checks on the `update-project-dependencies.sh` script

radtriste:
@evacchi @mbiarnes if you can have a look ? All scripts have been tested

PR #364, opened 2022-01-26
Setup PR job name for comment

radtriste:


Ginxo (approved):
LGTM. I just notice... Wouldn't it be nice to define initial variables (based on env variables) like we do here https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/734bcb6ecf28d4bce19ea11705a01340eeb9ea64/.ci/jenkins/Jenkinsfile.buildchain#L3 ?

This way it is easier to define which variables can be used for this particular pipeline, wdyt? We can include it as a part of a new PR. Thanks!

radtriste:
> LGTM. I just notice... Wouldn't it be nice to define initial variables (based on env variables) like we do here https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/734bcb6ecf28d4bce19ea11705a01340eeb9ea64/.ci/jenkins/Jenkinsfile.buildchain#L3 ?
> 
> This way it is easier to define which variables can be used for this particular pipeline, wdyt? We can include it as a part of a new PR. Thanks!

That would indeed give quite an overview. Gonna update that.

PR #377, opened 2022-02-01
[1.11.x] Install the build chain tool

rsynek:
There was one step completely missing, but it still might not be enough to fix the build.

### DIFF
--- a/.ci/actions/build-chain/action.yml
+++ b/.ci/actions/build-chain/action.yml
@@ -1,0 +1,45 @@
+name: 'Build Chain'
+description: 'Executes build-chain tool'
+inputs:
+  github-token:
+    description: "the github token"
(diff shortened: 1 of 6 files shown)


radtriste:
This will not work by itself as it also needs the `.ci/actions` to be present on the repository, aka on branch 1.11.x on kogito-pipelines
See https://github.com/kiegroup/kogito-pipelines/tree/main/.ci/actions

rsynek:
@radtriste is it ok to just copy-paste there what lies on the `main` branch, or can we expect any sort of incompatibilities?

radtriste:
The actions are not used in 1.11.x. So they will be there but unused. There should not be any problem.

radtriste (approved):

PR #399, opened 2022-03-04
PLANNER-2669 Disable full profile for GHA

rsynek:
https://issues.redhat.com/browse/PLANNER-2669

Related PRs:
https://github.com/kiegroup/optaplanner/pull/1871
https://github.com/kiegroup/optaplanner-quickstarts/pull/302
https://github.com/kiegroup/drools/pull/4245
https://github.com/kiegroup/kogito-pipelines/pull/399

### DIFF
--- a/.ci/pull-request-config.yaml
+++ b/.ci/pull-request-config.yaml
@@ -80,7 +80,7 @@ build:
   
   - project: kiegroup/optaplanner-quickstarts
     build-command: 
-      current: mvn clean install -f optaplanner-quickstarts/pom.xml -Dfull ${{ env.BUILD_MVN_OPTS }} ${{ env.BUILD_MVN_OPTS_CURRENT }} ${{ env.OPTAPLANNER_QUICKSTARTS_BUILD_MVN_OPTS }}
+      current: mvn clean install -f optaplanner-quickstarts/pom.xml ${{ env.BUILD_MVN_OPTS }} ${{ env.BUILD_MVN_OPTS_CURRENT }} ${{ env.OPTAPLANNER_QUICKSTARTS_BUILD_MVN_OPTS }}
       upstream: mvn clean install -Dquickly ${{ env.BUILD_MVN_OPTS }} ${{ env.BUILD_MVN_OPTS_UPSTREAM }} ${{ env..OPTAPLANNER_QUICKSTARTS_BUILD_MVN_OPTS_UPSTREAM }}
     clone:
       - optaplanner-quickstarts


.ci/pull-request-config.yaml
```diff
@@ -80,7 +80,7 @@ build:
   
   - project: kiegroup/optaplanner-quickstarts
     build-command: 
-      current: mvn clean install -f optaplanner-quickstarts/pom.xml -Dfull ${{ env.BUILD_MVN_OPTS }} ${{ env.BUILD_MVN_OPTS_CURRENT }} ${{ env.OPTAPLANNER_QUICKSTARTS_BUILD_MVN_OPTS }}
```
rsynek, line 83: I think this disabled the full profile for Jenkins builds as well. The question is, how to override it just for Jenkins?

radtriste: If I understand correctly, you want different PR checks for GHA and Jenkins ? not sure it should be done.
Main pupose of GHA was to replace Jenkins if that one is failing (backup PR check).

Well if you need to activate it only in Jenkins, you can setup the env variable `BUILD_MVN_OPTS_CURRENT` in upstream projects checking this one. which means:
- https://github.com/kiegroup/optaplanner/blob/main/.ci/jenkins/dsl/jobs.groovy#L51
- https://github.com/kiegroup/drools/blob/main/.ci/jenkins/dsl/jobs.groovy#L58
- https://github.com/kiegroup/optaplanner-quickstarts/blob/development/.ci/jenkins/dsl/jobs.groovy#L16

rsynek: The main reason is that building the distribution in both optaplanner and optaplanner-quickstarts takes a lot of time. For all the GHA matrix members. 
According to @triceo , on Windows, the optaplanner-quickstarts build takes ~50 minutes, on Linux ~30 minutes.

I see other reason why changing the behavior is not ideal: the more GHA and Jenkins differ, the more difficult it is to maintain them. However, we need to keep at least a single check that involves the distribution.

radtriste: we could make the GHA optional and activated by label only if Jenkins is unavailable. But that is another discussion.

Well then, you can keep this PRs to remove the full profile and update the 3 mentioned above files to have Jenkins specific configuration.

rsynek:
@radtriste when you have a moment, could you review this and the related PRs? Thanks!

radtriste (approved):

radtriste:
Sorry for the delay ...

PR #425, opened 2022-04-22
Correct Repo seeds

radtriste:


dsl/seed/jobs/scripts/seed_repo_generation.groovy
```diff
@@ -3,6 +3,38 @@ SEED_FOLDER = 'dsl/seed'
 util = null
 
 def generate() {
+    if ("${JOB_TYPE}" == 'TRIGGER') {
```
Ginxo, line 6: I'm not so confident about `==` usage, it's true that `TRIGGER` == "TRIGGER" => true, but I would prefer to use `equals` but up to you since this is something personal
```suggestion
    if ('TRIGGER'.equalsIgnoreCase("${JOB_TYPE}")) {
```

dsl/seed/src/main/groovy/org/kie/jenkins/jobdsl/SeedJobUtils.groovy
```diff
@@ -52,7 +55,11 @@ class SeedJobUtils {
                             }
                         }
                     }
-                    scriptPath('dsl/seed/jobs/Jenkinsfile.seed.trigger')
+                    if (repository == KogitoConstants.KOGITO_PIPELINES_REPOSITORY) {
```
Ginxo, line 58: it is the same here
```suggestion
                    if (KogitoConstants.KOGITO_PIPELINES_REPOSITORY.equalsIgnoreCase(repository)) {
```

Ginxo (approved):
LGTM, just a couple of suggestions

dsl/seed/jobs/scripts/seed_repo_generation.groovy
radtriste: well, in groovy, it calls the equals method: https://objectpartners.com/2018/11/07/equals-is-compareto-and-the-groovy-identity-operator/

radtriste: the rest of the code is using it, so I will let it. For string, it is not a big deal ;)

PR #450, opened 2022-05-18
[1.22.x] Fix pipelines for cloud images

radtriste:
Introduced by https://github.com/kiegroup/kogito-images/pull/1206

Related PRs:
- https://github.com/kiegroup/kogito-pipelines/pull/450
- https://github.com/kiegroup/kogito-images/pull/1262


dupliaka (commented):
LGTM maybe one suggestion if we leave function addAppsParam not in the shared library and implementation is a bit different for every usage how about changing the name to something more concrete?

radtriste:
I see your point. I named it to be in sync with `addExamplesParam`. I will rename them as part of the backport PR. Thank you for the suggestion !

PR #538, opened 2022-07-06
Hardcode seed branch in main configuration

radtriste:
This will ease the discovery of the branch and would avoid confusion


Ginxo (approved):
LGTM

PR #551, opened 2022-07-12
Add options to update-quarkus-platform script

radtriste:
- Disable checkout. In that case we need to be in the quarkus-platform for the script to work
- Setup the PR_BRANCH manually via parameter


mbiarnes (approved):
LGTM

evacchi (approved):

PR #587, opened 2022-08-19
[1.24.x] DSL: Disable triggers

radtriste:


mareknovotny:
cc @rgdoliveira @Ginxo @mbiarnes we should do it for prod pipelines too.

mareknovotny (approved):

PR #590, opened 2022-08-23
[BXMSPROD-1770] Centralize Operating Systems Configuration

lampajr:
JIRA:
- https://issues.redhat.com/browse/BXMSPROD-1770

referenced Pull Requests:
- https://github.com/kiegroup/kogito-pipelines/pull/590
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/2040

Improvements:

- Added OS Preparation Github action
- Allow triggering windows specific job using 'windows_check' label


lampajr:
Before **merging** the commit 30b50f200de5a5b47ec63e4306f9ca574fd91da6 must be reverted. It was added for testing purpose.

.ci/actions/os-preparation/matrix.json
```diff
@@ -0,0 +1,16 @@
+[
+  {
+    "os": ["ubuntu-latest"],
+    "event":"pull_request"
+  }, 
+  {
```
Ginxo: sorry for being picky here, but could we please follow the same order for all entries (os, event, filter) ?
```
  {
    "os": ["ubuntu-latest", "windows-latest"],
    "event":"pull_request",
    "repository": ["optaweb-vehicle-routing", "optaplanner", "optaplanner-quickstarts"]
  }
```

Ginxo (approved):
LGTM, just a minor thing

.ci/actions/os-preparation/action.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Check out kogito-pipelines
+      uses: actions/checkout@v3
+      with:
+        repository: lampajr/kogito-pipelines
+        ref: BXMSPROD-1770_os_preparation
+        path: matrix_preparation
```
radtriste: What will be the non testing values ?
You need to handle at least `main` and `release branches`.

radtriste: Maybe we should put some inputs with default values to the action:
- repository
- branch
- path

lampajr: > What will be the non testing values ? You need to handle at least `main` and `release branches`.

This would be the non testing config:
```
        repository: kiegroup/kogito-pipelines
        ref: main
        path: matrix_preparation
```



> Maybe we should put some inputs with default values to the action:
> 
> * repository
> * branch
> * path

That's a good point, in this case I would add only `owner` and `ref` parameter because the `path` one must not change since it is internally referenced, wdyt?

radtriste: well I just want to make it configurable. I could reuse that action to parse another file. Configuration may be different for Optaplanner and Kogito so it could be I need 2 different files

lampajr: yeah but the idea is to keep project-specific configuration centralized as well, just to give an example: suppose you want to add different os for a specific project you just need to add it here as:
```
{
    "os": ["ubuntu-latest", "windows-latest", "whatever"],
    "event":"pull_request",
    "repository": ["you-project"]
  }
```
In this case there is no need to change where the matrix.json configuration is taken, it will be always here. So the matrix will be always fetched from `kogito-pipelines` repo. Maybe I'm missing your specific use-case.

lampajr: btw I agree on making those parameters configurable, mainly for managing release branches in a custom way (if needed). Doing it.

Ginxo: I agree @radtriste we can define some inputs with default values for the os preparation as we already do for other composite actions, see https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/61fdbd4d6b666720f9987af7aea44af783b88318/.ci/actions/build-chain/action.yml#L3

radtriste: oh I see indeed. I just want to make it as flexible as possible as a composite action would be, for reusability :)
But that could be changed later if needed without impact I guess.

Also just thinking but, as the `matrix.json` file is a GHA file, I would expect it to be in the `.github` folder and not in the same folder as the composite action. Would make more sense IMO

lampajr: yeah it makes sense this too, it is much more appropriate under `.github` folder. Thanks a lot @radtriste !

.ci/actions/os-preparation/action.yaml
```diff
@@ -12,18 +22,24 @@ runs:
...
         repo_name: ${{ github.event.repository.name }}
         event: ${{ github.event_name }}
         label_name: ${{ github.event.label.name }}
+        config_file: matrix_preparation/.github/matrix-os.json
       shell: bash
       run: |
-        os=$(jq --arg repo "$repo_name" --arg event "$event" --arg event_value "$label_name" 'map(.| select(.event==$event and ((.value==null and ((.repository | index( $repo))) or (.repository==null)) or .value==$event_value)))' $GITHUB_WORKSPACE/matrix_preparation/.ci/actions/os-preparation/matrix.json | jq "if (length > 1) then (.[] | select(.repository!=null)) else (.[] | select(.repository==null)) end | .os")
-        echo "::set-output name=operating_systems::$(echo $os)"
+        os=$(jq --arg config_file "$config_file" --arg repo "$repo_name" --arg event "$event" --arg event_value "$label_name" 'map(.| select(.event==$event and ((.value==null and ((.repository | index( $repo))) or (.repository==null)) or .value==$event_value)))' $GITHUB_WORKSPACE/$config_file | jq "if (length > 1) then (.[] | select(.repository!=null)) else (.[] | select(.repository==null)) end | .os")
+        if [ -z "$os"]; then
```
radtriste:
```suggestion
        if [ -z "$os" ]; then 
```

.github/matrix-os.json
```diff
@@ -0,0 +1,16 @@
...
+  {
+    "os": ["ubuntu-latest", "windows-latest"],
+    "event":"pull_request",
+    "repository": ["optaweb-vehicle-routing", "optaplanner", "optaplanner-quickstarts"]
+  },
+  {
+    "os": ["windows-2022"],
+    "event":"label",
+    "value":"windows_check"
+  }
```
radtriste, line 15: does that event is retriggered when there is a new commit ?

lampajr: no, it should be triggered only on labeling the PR - a commit should be a 'pull_request' event even if a label is applied.

radtriste (approved):
I am ok with this first version. We can improve it later on if needed

PR #604, opened 2022-09-06
Branch seed job should use main seed branch

radtriste:
if no seed branch is defined


dsl/seed/jenkinsfiles/Jenkinsfile.seed.main
```diff
@@ -172,6 +172,10 @@ String getBranchSeedJobName() {
     return '0-seed-job'
 }
 
+boolean isSeedBranchExisting(String branch) {
```
rsynek: Why is the method empty and not used?

radtriste: I started some dev but realize this was useless and there was an easiest method :D
I just forgot to remove this ... done

rsynek (approved):

PR #605, opened 2022-09-07
KOGITO-7076 Setup maven settings in composite action

radtriste:
We need to make sure that the `kiegroup` snapshot artifacts are not downloaded from Nexus to avoid any compatibility issue ...
This is already the case on Jenkins, thus we are sometimes having differences in behavior ...
At least now, the `settings.xml` file will be similar for PRs.

### DIFF
--- a/.ci/actions/maven/action.yml
+++ b/.ci/actions/maven/action.yml
@@ -12,6 +12,10 @@ inputs:
   cache-key-prefix:
     description: "the cache key"
     required: false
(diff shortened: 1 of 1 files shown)


radtriste:
Tested in https://github.com/kiegroup/drools/pull/4669 and https://github.com/kiegroup/optaplanner/pull/2098

Ginxo (approved):
LGTM

PR #680, opened 2022-11-07
[KOGITO-6259] enabled trigger

mbiarnes:


radtriste:
@mbiarnes I think there was a misunderstanding on the issue purpose :)
The issue is to create a job which would allow to disable the triggers on the release branch where it executes ;)

PR #752, opened 2022-12-16
[BXMSPROD-1936] Fix default reviewers and customization

lampajr:
**JIRA**: https://issues.redhat.com/browse/BXMSPROD-1936

**referenced Pull Requests**:

* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/2175
* https://github.com/kiegroup/kogito-pipelines/pull/752

**Improvements**

Added new input parameters:
* `include-merge-as-reviewer`: whether to include or not the user who merged the PR as reviewer, default `true`
* `include-requested-reviewers`: whether to include original requested reviewers into backported pr, notice that approved reviewers are not in this list since ti seems github remove them. Default `true`
* `custom-reviewers`: allows users to customize the workflow adding additional custom reviewers

Deprecated `additional-reviewers` as it is now managed internally in the composite actions itself, this field will be ignored since now.

Thus we can remove that parameter from any workflow without changing the final result as `include-requested-reviewers` is set to true by default.

### DIFF
--- a/.ci/actions/backporting/action.yml
+++ b/.ci/actions/backporting/action.yml
@@ -25,10 +25,21 @@ inputs:
     description: "Author of the original pull request"
     default: "${{ github.event.pull_request.user.login }}"
     required: false
-  additional-reviewers:
-    description: "Additional backported pull request reviewers, e.g., toJSON(github.event.pull_request.requested_reviewers)"
-    default: "[]"
+  include-merge-as-reviewer:
+    description: "Whether include or not who merged the pull request as reviewer of the backported one"
+    default: "true"
     required: false
+  include-requested-reviewers:
+    description: "Whether include or not original requested reviewers"
+    default: "true"
+    required: false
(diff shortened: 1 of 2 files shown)


lampajr:
@radtriste this time it seems that getting the github context from the composite action worked (most likely I did something wrong last time), btw here is my proposal that should make the actions itself much more customizable. Furthermore all defaults are setup directly here instead of having to provide some additional inputs from workflows. wdyt?

radtriste (approved):
lgtm

lampajr:
fyi @mareknovotny @Ginxo if you agree too, I'll propagate the same change on droolsjbpm-build-boostrap

mareknovotny (approved):

lampajr:
Upgraded `github-cherry-pick-action` to `1.0.6` that simply upgrade node to `node16`

Ginxo (approved):
LGTM

PR #765, opened 2022-12-23
KOGITO-8090 GHA Maven cache exclude kie artifacts

radtriste:


danielezonca:
Good idea, it is way safer :)

Is it possible also make sure that all snapshots are excluded from cache?

radtriste:
> Good idea, it is way safer :)
> 
> Is it possible also make sure that all snapshots are excluded from cache?

I would say that this is even better, just ignoring SNAPSHOT folders/files

radtriste:
@lampajr @danielezonca @Ginxo See https://github.com/kiegroup/kogito-pipelines/actions/runs/3969019483 for example

lampajr (approved):
LGTM

danielezonca (approved):

radtriste:
@lampajr if we are good, then we can merge that one :)

PR #792, opened 2023-01-27
KOGITO-8539 Correct quarkus integration check

radtriste:
https://issues.redhat.com/browse/KOGITO-8539
Should create the integration branch only for the current project and not for upstream


lampajr (approved):
LGTM

PR #827, opened 2023-02-06
Prevent action-surefire-report from creating a new check run

shubhbapna:


shubhbapna:
merging since there is no other way to test it

PR #942, opened 2023-04-19
[BXMSPROD-728] add action-junit-report

shubhbapna:
JIRA: https://issues.redhat.com/browse/BXMSPROD-728

While we can't use checks run API since GitHub has a bug in their API (reported [here](https://github.com/orgs/community/discussions/14891) and [here](https://github.com/orgs/community/discussions/24616)), we can add annotations and produce a job summary that will indicate test failures. We will be able to achieve easy distinction between test failures and build failures as soon as we click on the checks that a PR ran.

### DIFF
--- a/.ci/actions/action-junit-report/action.yml
+++ b/.ci/actions/action-junit-report/action.yml
@@ -1,0 +1,25 @@
+name: 'Action Junit Report'
+description: 'Generates a test report and annotates failures'
+inputs:
+  report_paths:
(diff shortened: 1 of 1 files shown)


shubhbapna:
@radtriste if we are not using surefire-report anywhere else do you want me to remove that action?

mareknovotny (approved):

radtriste (approved):

PR #943, opened 2023-04-19
[BXMSPROD-2018] Setup kogito status page

lampajr:
**JIRA**: https://issues.redhat.com/browse/BXMSPROD-2018

This PR aims to setup Kogito status page like the one we have for Kiegroup https://kiegroup.github.io/droolsjbpm-build-bootstrap/status/kiegroup-status.

Here main configuration:

- Project list is taken from https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/.ci/pull-request-config.yaml 
```
definition-file: https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/.ci/pull-request-config.yaml
```

- Take into account the following branches when providing PRs infos
```
base-branch-filter: main,1.13.x,1.13.x-blue,8.13.x
```

- Provide branches comparison for the following two branches
```   
branches: 1.13.x,1.13.x-blue
```

### DIFF
--- a/.github/workflows/generate_status_page_data.yaml
+++ b/.github/workflows/generate_status_page_data.yaml
@@ -1,0 +1,28 @@
+name: Generate status page data
+
+on:
+  workflow_dispatch:
+    schedule:
+      - cron: '0 * * * *'
+
+jobs:
+  generate-status-page-data:
+    if: github.repository_owner == 'kiegroup'
+    concurrency:
+      group: generate-status-page-data
+      cancel-in-progress: true
+    runs-on: ubuntu-latest
+    name: Generate status page data
+    steps:
+      - name: Generate status page data
+        uses: kiegroup/chain-status/.ci/actions/generate-data@main
+        with:
+          definition-file: https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/.ci/pull-request-config.yaml
+          title: Kogito Status
+          subtitle: Kiegroup Kogito organization repositories CI status
+          base-branch-filter: main,1.13.x,1.13.x-blue,8.13.x
+          branches: 1.13.x,1.13.x-blue
+          created-by: GitHub Action
+          created-url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          logger-level: debug
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/generate_status_page_data.yaml
+++ b/.github/workflows/generate_status_page_data.yaml
@@ -14,13 +14,26 @@ jobs:
     runs-on: ubuntu-latest
     name: Generate status page data
     steps:
+      - uses: actions/checkout@v3
+        with:
+          path: ./kogito-pipelines
+      - name: Retrieve active repositories
+        id: retrieve_active_repos
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq '[.repositories[].name] | join(",")' ./kogito-pipelines/dsl/config/branch.yaml
+      - name: Retrieve active branches
+        id: retrieve_active_branches
+        uses: mikefarah/yq@master
+        with:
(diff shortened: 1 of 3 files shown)


lampajr:
> - Project list is taken from https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/.ci/pull-request-config.yaml 
```
definition-file: https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/.ci/pull-request-config.yaml
```

Regarding this I noticed that the bc config does not cover all projects we have on Kogito 1.13.x* because the main branch now just contains those for OSL. On the opposite if we try to use bc config from 1.13.x we will miss some projects used by OSL (like kogito-apps).

Therefore I think the only way to include all of them would be to replace `definition-file` param with `projects` [1] providing a custom list of projects we want to include (it whould not be too long), wdyt?

[1] https://github.com/kiegroup/chain-status/blob/main/.ci/actions/generate-data/action.yml#L10

radtriste:
well in fact, you would need to create the same PR on Drools and OP main branches and use the different build-chain config: that would mean 3 different pages: kogito, drools and OP
you could also look into the `branch.yaml` config file on each repository.

.github/workflows/generate_status_page_data.yaml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    name: Generate status page data
+    steps:
+      - name: Generate status page data
+        uses: kiegroup/chain-status/.ci/actions/generate-data@main
+        with:
+          definition-file: https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/.ci/pull-request-config.yaml
+          title: Kogito Status
+          subtitle: Kiegroup Kogito organization repositories CI status
+          base-branch-filter: main,1.13.x,1.13.x-blue,8.13.x
```
radtriste: why not using the "active" branches defined into the `main.yaml` config file ?
that way you keep the dynamic in case of any change :)

lampajr: @radtriste that is avery good idea, as you suggested we could use:
- [main.yaml](https://github.com/kiegroup/kogito-pipelines/blob/main/dsl/config/main.yaml#L19) --> retrieve active branches 
- [branch.yaml](https://github.com/kiegroup/kogito-pipelines/blob/main/dsl/config/branch.yaml#L70) --> retrieve active repositories

lampajr:
In commit https://github.com/kiegroup/kogito-pipelines/pull/943/commits/2ca2fb76a34afa9b4da56477dbb2df6a3888e5bc I added the following changes:
- Two new steps to retrieve active branches and repositories from DSL config
- Generate status page data from projects' list instead of defintion file

.github/workflows/generate_status_page_data.yaml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          cmd: yq '[.git.branches[].name] | join(",")' ./kogito-pipelines/dsl/config/main.yaml
+      - name: Generate status page data
+        uses: kiegroup/chain-status/.ci/actions/generate-data@main
+        with:
+          projects: "${{ steps.retrieve_active_repos.outputs.result }}"
+          title: Kogito Status
+          subtitle: Kiegroup Kogito organization repositories CI status
+          base-branch-filter: "${{ steps.retrieve_active_branches.outputs.result }}"
+          branches: 1.13.x,1.13.x-blue
```
radtriste, line 37: What is the difference between the 2 ?

radtriste (approved):
just one question

.github/workflows/generate_status_page_data.yaml
lampajr: Basically
`base-branch-filter`:  used to filter PRs that have as base branch one of those defined in this variable
`branches`: for those branches listed here, the tool retrieves additional information that is used later to provide branches comparison

I know that is not so self-explainatory :(

mareknovotny (approved):

Ginxo (approved):
LGTM

PR #946, opened 2023-04-24
[BXMSPROD-2019] Setup automated prod branch creation

lampajr:
**JIRA:** https://issues.redhat.com/browse/BXMSPROD-2019

### DIFF
--- a/dsl/config/branch.yaml
+++ b/dsl/config/branch.yaml
@@ -69,9 +69,13 @@ disable:
   triggers: false
 repositories:
 - name: kogito-pipelines
+  to_productize: true
 - name: kogito-runtimes
+  to_productize: true
 - name: kogito-apps
+  to_productize: true
 - name: kogito-examples
+  to_productize: true
 - name: kogito-images
 - name: kogito-operator
 - name: kogito-serverless-operator
@@ -81,7 +85,8 @@ repositories:
   branch: main-kogito
 git:
   author:
-    name: kiegroup
+    # TODO change back to kiegroup
+    name: lampajr
     credentials_id: kie-ci3
     token_credentials_id: kie-ci3-token
   bot_author:
--- a/dsl/config/branch.yaml
+++ b/dsl/config/branch.yaml
@@ -69,13 +69,9 @@ disable:
   triggers: false
 repositories:
 - name: kogito-pipelines
-  to_productize: true
 - name: kogito-runtimes
-  to_productize: true
 - name: kogito-apps
-  to_productize: true
 - name: kogito-examples
-  to_productize: true
 - name: kogito-images
 - name: kogito-operator
 - name: kogito-serverless-operator
@@ -83,10 +79,16 @@ repositories:
 - name: kogito-docs
 - name: kie-docs
   branch: main-kogito
+productized_repositories:
+- name: kogito-pipelines
+- name: kogito-runtimes
+- name: kogito-apps
+- name: kogito-examples
+- name: kogito-images
+- name: kogito-operator
 git:
   author:
-    # TODO change back to kiegroup
-    name: lampajr
+    name: kiegroup
     credentials_id: kie-ci3
     token_credentials_id: kie-ci3-token
   bot_author:
--- a/dsl/config/branch.yaml
+++ b/dsl/config/branch.yaml
@@ -85,7 +85,7 @@ productized_repositories:
 - name: kogito-apps
 - name: kogito-examples
 - name: kogito-images
-- name: kogito-operator
+- name: kogito-serverless-operator
 git:
   author:
     name: kiegroup
--- a/dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
+++ b/dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
@@ -1,0 +1,360 @@
+import org.jenkinsci.plugins.workflow.libs.Library
+import java.util.regex.Pattern
+
+@Library('jenkins-pipeline-shared-libraries')_
+
+// TODO: add parameters
+// 1. <prj>_RELEASE_BRANCH
+// 2. QUARKUS_VERSION
+// 3. 
+
+seedConfig = [:]
+mainBranchConfig = [:]
+repositoriesToBranch = []
+
+pipeline{
+    agent{
+        label 'kie-rhel7 && !built-in'
+    }
+    stages{
+        stage('Initialize'){
+            steps{
+              script {
+                    checkout scm
+
+                    String projectListStr = getEcosystemProjects().collect { "${getRepoNameCamelCase(it.name)} ${getBranchFromRepository(it.name)}" }.join(' / ')
+                    currentBuild.displayName = "Productize Branch for ${projectListStr}"
+              }
+            }
+        }
+
+        stage('Create productized branch'){
+            steps{
+                script {
+                    List repositories = getRepositoriesToBranch()
+                    createBranches(repositories)
+                }
+            }
+        }
+
+        // update the main seed config by adding the new prod branch, e.g.,
+        //   - name: 8.36.x-prod
+        //      seed:
+        //      branch: seed-drools-8.36.x // this branch MUST already exist
+        stage('Update main seed config') {
+            steps {
+                script {
+                    // String newBranch = getBranchConfigReleaseBranch() // FIXME: remove since not needed
+                    String seedConfigFile = env.SEED_CONFIG_FILE_PATH
+
+                    dir(checkoutMainSeedConfigFileRepo()) {
+                        def seedConfig = readYaml file: seedConfigFile
+
+                        if (!seedConfig.git.branches.find { it.name == getEcosystemProductizedBranch() }) {
+                            def branchConfig = [
+                                name: getEcosystemProductizedBranch(),
+                                seed: [
+                                    branch: getSeedRepoReleaseBranch(),
+                                ]
+                            ]
+                            // FIXME: remove? no need I think 
+                            // if (newBranch != getEcosystemReleaseBranch()) {
+                            //     branchConfig.seed.config_file = [
+                            //         git: [
+                            //             branch: newBranch
+                            //         ]
+                            //     ]
+                            // }
+                            seedConfig.git.branches.add(branchConfig)
+                        }
+
+                        echo "Write seed config => ${seedConfig}"
+                        sh "rm ${seedConfigFile}"
+                        writeYaml file: seedConfigFile, data: seedConfig, overwrite: true
+
+                        pushIfChanged(env.SEED_CONFIG_FILE_GIT_BRANCH, env.SEED_CONFIG_FILE_GIT_AUTHOR_CREDS_ID, "Add productized branch ${getEcosystemProductizedBranch()} after upstream release")
+                    }
+                }
+            }
+        }
+
+        // 1. update productized branch seed config repo
+        //          this will:
+        //              - disable those jobs that are no more needed on branch.yaml [this MUST be executed only on seed repo, e.g., kogito-pipelines]
+        //              - enable productized profile (-Dproductized) on jenkins environments branch.yaml [this MUST be executed only on seed repo, e.g., kogito-pipelines]
+        //              - disable some environments when branching as they make no sense by default [this MUST be executed only on seed repo, e.g., kogito-pipelines]
+        //              - create PR ? or do these changes directly on the prod branch
+
+        stage("Update prod branch seed config repo"){
+            steps{
+               script {
+                // TODO: implement
+               }
+            }
+        }
+
+        // 2. update productized branch on other projects
+        //              - disable PR checks from jobs.groovy
+        //              - comment out projects we are not interested in from the buildchain-config.yaml
+        //              - disable GHA jobs no more needed by commenting out workflows
+        //              - enable productized profile (-Dproductized) on GHA PR checks workflows
+        //              - ensure that fullProfile is not enabled on .ci/jenkins/dsl/jobs.groovy, if so remove it
+
+        stage("Update prod branch for all projects to be productized"){
+            steps{
+               script {
+                // TODO: implement
+               }
+            }
+        }
+
+        // 3. call update quarkus job in order to downgrade quarkus to LTS version (version provided as parameter QUARKUS_VERSION)
+        //          this will:
+        //              - downgrade quarkus, there already exists a job doing this
+        //              - apply patches from .ci/environments/quarkus-lts on the same PR
+
+        stage('Downgrade quarkus to LTS version'){
+            steps{
+                script {
+                    // TODO: implement
+                }
+            }
+        }
+    }
+    post{
+        unsuccessful {
+            script {
+                // TODO: send notification
+            }
+        }
+        success {
+            script {
+                // TODO: send notification with new branches and PRs if any
+            }
+        }
+    }
+}
+
+// Retrieve job parameters keys
+
+String getBranchParameterKey(String projectName) {
+    return "${projectName}_RELEASE_BRANCH".toUpperCase()
+}
+
+String getProdBranchSuffixParameterKey(String projectName) {
+    return "${projectName}_PROD_BRANCH_SUFFIX".toUpperCase()
+}
+
+String getQuarkusVersionParameterKey() {
+    return "QUARKUS-VERSION"
+}
+
+// Retrieve seed configuration
+
+String checkoutMainSeedConfigFileRepo() {
+    assert env.SEED_CONFIG_FILE_GIT_REPOSITORY : 'Missing main seed config file Git Repository into the environment. Please check the job configuration.'
+    assert env.SEED_CONFIG_FILE_GIT_AUTHOR_NAME : 'Missing main seed config file Git Author Name into the environment. Please check the job configuration.'
+    assert env.SEED_CONFIG_FILE_GIT_AUTHOR_CREDS_ID : 'Missing main seed config file Git Author Credentials ID into the environment. Please check the job configuration.'
+    assert env.SEED_CONFIG_FILE_GIT_BRANCH : 'Missing main seed config file Git Branch into the environment. Please check the job configuration.'
+
+    String dirName = "seed-main-${env.SEED_CONFIG_FILE_GIT_REPOSITORY}"
+    dir(dirName) {
+        deleteDir()
+        checkout(githubscm.resolveRepository(env.SEED_CONFIG_FILE_GIT_REPOSITORY, env.SEED_CONFIG_FILE_GIT_AUTHOR_NAME, env.SEED_CONFIG_FILE_GIT_BRANCH, false, env.SEED_CONFIG_FILE_GIT_AUTHOR_CREDS_ID))
+        sh "git checkout ${env.SEED_CONFIG_FILE_GIT_BRANCH}"
+    }
+    return dirName
+}
+
+def readSeedConfig() {
+    assert env.SEED_CONFIG_FILE_PATH : 'Missing main seed config file Path into the environment. Please check the job configuration.'
+    if (!seedConfig) {
+        dir(checkoutMainSeedConfigFileRepo()) {
+            seedConfig = readYaml(file: "${env.SEED_CONFIG_FILE_PATH}")
+        }
+    }
+    return seedConfig
+}
+
+String getMainBranchConfigFilePath() {
+    return readSeedConfig().seed.config_file.path
+}
+
+String getMainBranchConfigFileGitAuthorCredentialsId() {
+    return readSeedConfig().seed.config_file.git.author.credentials_id
+}
+
+String checkoutMainBranchSeedConfigFileRepo() {
+    String dirName = "seed-branch-${getMainBranchConfigFileGitRepository()}"
+    dir(dirName) {
+        deleteDir()
+        checkout(githubscm.resolveRepository(getMainBranchConfigFileGitRepository(), getMainBranchConfigFileGitAuthorName(), getMainBranchConfigFileGitBranch(), false, getMainBranchConfigFileGitAuthorCredentialsId()))
+        sh "git checkout ${getMainBranchConfigFileGitBranch()}"
+    }
+    return dirName
+}
+
+def readMainBranchConfig() {
+    if (!mainBranchConfig) {
+        dir(checkoutMainBranchSeedConfigFileRepo()) {
+            mainBranchConfig = readYaml(file: "${getMainBranchConfigFilePath()}")
+        }
+    }
+    return mainBranchConfig
+}
+
+String getRepoNameCamelCase(String repo) {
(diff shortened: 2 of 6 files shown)


dsl/config/branch.yaml
```diff
@@ -81,7 +85,8 @@ repositories:
   branch: main-kogito
 git:
   author:
-    name: kiegroup
+    # TODO change back to kiegroup
```
radtriste: test commit example that I usually cherry-pick on my PR branch: https://github.com/radtriste/kogito-pipelines/commit/185b5b2e2c42b62f76bdfcd8c8c45957f1e5e529
I just update then the `test-config` branch to the correct branch

dsl/config/branch.yaml
```diff
@@ -69,9 +69,13 @@ disable:
   triggers: false
 repositories:
 - name: kogito-pipelines
+  to_productize: true
```
radtriste: instead of having `to_productize` in that file and `ignore_productization` in the other, why not having simply a `productized_repositories` key at the root here ?

dsl/seed/jobs/root_jobs.groovy
```diff
@@ -84,3 +84,44 @@ if (nonMainBranches) {
...
+]
+
+KogitoJobTemplate.createPipelineJob(this, jobParamsProd)?.with {
+    parameters {
+        PRODUCTIZED_PROJECTS.split(',').each { projectName ->
+            stringParam("${projectName}_RELEASE_BRANCH".toUpperCase(), '', "${Utils.getRepoNameCamelCase(projectName)} version to release as Major.minor.micro")
+        }
+
+        stringParam('QUARKUS_VERSION', '', 'Quarkus version to which to update all productized branches, usually latest LTS version')
+        booleanParam('PRODUCTIZED_BRANCH', true, 'Is the created branch a productized one ?')
```
radtriste: do we need that one ?

dsl/config/branch.yaml
lampajr: yeah that makes totally sense

radtriste: oh I see, maybe keep the `ignore_productization`
but I think it would be clearer to have in the branch.yaml a `productized_repositories` list which would match some of the `repositories`
btw, you may also want to productize the branches for kogito-images and operator or ?

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,360 @@
...
+// 2. QUARKUS_VERSION
+// 3. 
+
+seedConfig = [:]
+mainBranchConfig = [:]
+repositoriesToBranch = []
+
+pipeline{
+    agent{
+        label 'kie-rhel7 && !built-in'
```
radtriste: we should drop the `kie-rhel7` and use the `rhel8` label instead

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,360 @@
...
+    }
+    stages{
+        stage('Initialize'){
+            steps{
+              script {
+                    checkout scm
+
+                    String projectListStr = getEcosystemProjects().collect { "${getRepoNameCamelCase(it.name)} ${getBranchFromRepository(it.name)}" }.join(' / ')
```
radtriste, line 25: you may want only the productized ones here or ?

dsl/config/branch.yaml
lampajr: yeah I would keep `ignore_productization` on main.yaml, but I agree on having a separated list of productized repositories.. it will be clearer.
yes I will add images and operator too, this was just an initial setup I did not pay attention on true repos to include atm

dsl/seed/jobs/root_jobs.groovy
```diff
@@ -84,3 +84,44 @@ if (nonMainBranches) {
...
+        author: Utils.getSeedAuthor(this),
+        credentials: Utils.getSeedAuthorCredsId(this),
+        branch: Utils.getSeedBranch(this),
+    ],
+    env: [:],
+    jenkinsfile: 'dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare',
+]
+
+KogitoJobTemplate.createPipelineJob(this, jobParamsProd)?.with {
+    parameters {
```
radtriste: in parameters, we should have a `BASE_BRANCH` parameter which would be a `choiceParameter` which list all the available community branches.
Then we should create the prod branch from this community branch.
See also https://github.com/kiegroup/kogito-pipelines/blob/main/dsl/seed/jobs/root_jobs.groovy#L72 as an example

dsl/seed/jobs/root_jobs.groovy
lampajr: well actually not as parameter as it should be always true

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
lampajr: correct, that is what I am doing.. `getEcosystemProjects` should returns just the projects that we would like to productize, i.e., ignoring those that have `ignore_productization` set to `true`

dsl/seed/jobs/root_jobs.groovy
lampajr: yeah that was my idea but I was thinking of following the same approach you used for release process.. i.e., having one branch per project named like `<REPO>_RELEASE_BRANCH` that will be used as base branch. I already create those params.. I just need to convert them to be `choiceParameter`

dsl/seed/jobs/root_jobs.groovy
lampajr: done

dsl/config/branch.yaml
lampajr: done

dsl/seed/jobs/root_jobs.groovy
lampajr: done: updated to `choiceParam` but keeping one param per repo

dsl/seed/jobs/root_jobs.groovy
```diff
@@ -84,3 +84,48 @@ if (nonMainBranches) {
...
+    env: [:],
+    jenkinsfile: 'dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare',
+]
+
+
+if (nonMainBranches) {
+    KogitoJobTemplate.createPipelineJob(this, jobParamsProd)?.with {
+        parameters {
+            PRODUCTIZED_PROJECTS.split(',').each { projectName ->
+                choiceParam("${projectName}_RELEASE_BRANCH".toUpperCase(), nonMainBranches, "${Utils.getRepoNameCamelCase(projectName)} community branch to which to create the productized branch from")
```
radtriste: NTH: should we eliminate the `*-prod` from the `nonMainBranches` ?

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,358 @@
...
+String getEcosystemProductizedBranch() {
+    return getProductizedBranchFromProject(getEcosystemMainProject())
+}
+
+String getEcosystemReleaseBranch() {
+    return getBranchFromProject(getEcosystemMainProject())
+}
+
+String getSeedRepoReleaseBranch() {
+    return "seed-${getEcosystemMainProject()}-${getEcosystemReleaseBranch()}"
```
radtriste: This should be read from the `main.yaml` file, based on the `getEcosystemReleaseBranch`. This would be more accurate instead of "generating" it
That way you make sure you use the seed branch existing

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,358 @@
...
+                        def seedConfig = readYaml file: seedConfigFile
+
+                        if (!seedConfig.git.branches.find { it.name == getEcosystemProductizedBranch() }) {
+                            def branchConfig = [
+                                name: getEcosystemProductizedBranch(),
+                                seed: [
+                                    branch: getSeedRepoReleaseBranch(),
+                                ]
+                            ]
+                            // FIXME: remove? no need I think
```
radtriste: yes can be removed

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        // update the main seed config by adding the new prod branch, e.g.,
+        //   - name: 8.36.x-prod
+        //      seed:
+        //      branch: seed-drools-8.36.x // this branch MUST already exist
+        stage('Update main seed config') {
+            steps {
+                script {
+                    // String newBranch = getBranchConfigReleaseBranch() // FIXME: remove since not needed
```
radtriste: yes can be removed

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,358 @@
...
+    return "${projectName}_RELEASE_BRANCH".toUpperCase()
+}
+
+String getProdBranchSuffixParameterKey(String projectName) {
+    return "${projectName}_PROD_BRANCH_SUFFIX".toUpperCase()
+}
+
+String getQuarkusVersionParameterKey() {
+    return "QUARKUS-VERSION"
+}
```
radtriste:
```suggestion
String getQuarkusVersion() {
     return params.QUARKUS_VERSION
}
```

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,358 @@
...
+                    List repositories = getRepositoriesToBranch()
+                    createBranches(repositories)
+                }
+            }
+        }
+
+        // update the main seed config by adding the new prod branch, e.g.,
+        //   - name: 8.36.x-prod
+        //      seed:
+        //      branch: seed-drools-8.36.x // this branch MUST already exist
```
radtriste, line 43: and it should, see my other comment on `getSeedRepoReleaseBranch` method

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,358 @@
...
+                    }
+                }
+            }
+        }
+
+        // 1. update productized branch seed config repo
+        //          this will:
+        //              - disable those jobs that are no more needed on branch.yaml [this MUST be executed only on seed repo, e.g., kogito-pipelines]
+        //              - enable productized profile (-Dproductized) on jenkins environments branch.yaml [this MUST be executed only on seed repo, e.g., kogito-pipelines]
+        //              - disable some environments when branching as they make no sense by default [this MUST be executed only on seed repo, e.g., kogito-pipelines]
```
radtriste: not on seed repo but on config files repository, which would be retrieved from `main.yaml` config
this branch config file can be retrieved the same way as `readMainBranchConfig()` and `getMainBranchConfigFile[...]` methods in https://github.com/kiegroup/kogito-pipelines/blob/main/dsl/seed/jenkinsfiles/Jenkinsfile.release.prepare
but with the correct branch (instead of main branch)

radtriste: in our testing example, this would be `kogito-pipelines:8.36.x-prod`

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,358 @@
...
+        //              - create PR ? or do these changes directly on the prod branch
```
radtriste: you should directly push the changes IMO (or create PR in the first run to be sure changes are ok)

radtriste (commented):
another round :)

dsl/seed/jobs/root_jobs.groovy
lampajr: good catch, you are right as at some point we will have also those one on the config

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
lampajr: yeah that makes sense but only if you are productizing a branch that is present on main.yaml which is not always true (or better we cannot guarantee that - e.g., consider the case where we would like to productize an old branch that is no more present on the main.yaml) - so I was thinking about gathering that information as you suggested and fallback to this creation only if not found, wdyt @radtriste  ?

radtriste: that would be perfect

lampajr:
In order to enable prod profile is not always enough to add `-Dproductized` as for some projects we might have to add additional params, for this goal I added a new string param `ADDITIONAL_BUILD_MAVEN_OPTS` where we can specifiy additional maven options before running the job. wdyt?

https://github.com/kiegroup/kogito-pipelines/pull/946/files#diff-d236b73b7f1d4d4792f498847645661bc46bf8f614c1482a33ee706f359f3316R113-R120

rgdoliveira:
> In order to enable prod profile is not always enough to add `-Dproductized` as for some projects we might have to add additional params, for this goal I added a new string param `ADDITIONAL_BUILD_MAVEN_OPTS` where we can specifiy additional maven options before running the job. wdyt?

I think it is good idea, just out of curiosity, are the same parameters valid/acceptable by all the projects? Or some projects require specific parameters that others project can't have?

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,416 @@
...
+        stage("Update branch config on productized branch"){
+            steps{
+                script {
+                    def prodBranch = getBranchConfigProductizedBranch()
+                    def prodRepositories = getProdRepositoriesToBranch()
+
+                    dir(checkoutMainBranchSeedConfigFileRepo()) {
+                        getOrCreateGitBranch(prodBranch, getMainBranchConfigFileGitAuthorCredentialsId())
+
+                        def branchConfigFilePath = getMainBranchConfigFilePath()
```
radtriste: getProdBranchConfigFilePath ?

lampajr:
> > In order to enable prod profile is not always enough to add `-Dproductized` as for some projects we might have to add additional params, for this goal I added a new string param `ADDITIONAL_BUILD_MAVEN_OPTS` where we can specifiy additional maven options before running the job. wdyt?
> 
> I think it is good idea, just out of curiosity, are the same parameters valid/acceptable by all the projects? Or some projects require specific parameters that others project can't have?

Well right now the additional parameter that is needed is `-Ddata-index-ephemeral.image=quay.io/kiegroup/kogito-data-index-ephemeral` that maybe can be added on all projects and it will be ignored if not really used by a project

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,416 @@
...
+                        // disable some environments when branching as they make no sense by default
+                        branchConfig.environments.each { envName, envData ->
+                            if (envName.startsWith('quarkus')) {
+                                if (envName != 'quarkus-branch') {
+                                    envData.enabled = false // Keep only quarkus-branch
+                                }
+                            }
+                            if (envData.ids.find { it == 'lts' }) {
+                                envData.enabled = branchConfig.productized_branch
+                            }
```
radtriste: prod branch will already be LTS, so I don't think we need to enable those

In the end, the env which will be important for prod branch will be:
- native
- quarkus-branch

wdyt ?

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,416 @@
...
+                        }
+
+                        // disable DSL triggers
+                        branchConfig.disable.triggers = true
```
radtriste: we don't want any community run ?
If you disable triggers, then you will have no nightlies in community. Is that expected behavior ?

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,416 @@
...
+
+                        // enable productized profile
+                        def buildMvnOpts = "-Dproductized ${getAdditionalDefaultBuildMavenOpts()} "
+
+                        branchConfig.environments.each { envName, envData ->
+                            if (envData.enabled && envData.env_vars.BUILD_MVN_OPTS) {
+                                envData.env_vars.BUILD_MVN_OPTS = buildMvnOpts + envData.env_vars.BUILD_MVN_OPTS
+                            }
+                        }
+                        branchConfig.default.env_vars.BUILD_MVN_OPTS = buildMvnOpts
```
radtriste:
```suggestion
                        branchConfig.environments.default.env_vars.BUILD_MVN_OPTS = buildMvnOpts
```

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,416 @@
...
+                        branchConfig.disable.triggers = true
+
+
+                        // enable productized profile
+                        def buildMvnOpts = "-Dproductized ${getAdditionalDefaultBuildMavenOpts()} "
+
+                        branchConfig.environments.each { envName, envData ->
+                            if (envData.enabled && envData.env_vars.BUILD_MVN_OPTS) {
+                                envData.env_vars.BUILD_MVN_OPTS = buildMvnOpts + envData.env_vars.BUILD_MVN_OPTS
+                            }
```
radtriste: here if `envData.env_vars.BUILD_MVN_OPTS` does not exist, you don't add any prod profile ?

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
radtriste: what if the default env already exists ?
if would be the same behavior as in other envs IMO

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
lampajr: oh yeah I think you are right, in this way quarkus-branch and native jobs are actually the LTS ones

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
lampajr: yeah you are right, I used the same approach we used for drools 8.36.x-prod but that was not a "true" productized branch.. that's why we did that. Fixed now.

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
lampajr: I applied the same change while iterating over all envs, as you said no need to do this separately

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,407 @@
...
+                                }
+                            }
+                        }
+
+
+                        // enable productized profile
+                        def buildMvnOpts = "-Dproductized ${getAdditionalDefaultBuildMavenOpts()} "
+
+                        branchConfig.environments.each { envName, envData ->
+                            envData.env_vars.BUILD_MVN_OPTS = buildMvnOpts + (envData.env_vars.BUILD_MVN_OPTS ?: '')
```
radtriste, line 113:
```suggestion
                            envData.env_vars = envData.env_vars ?: [:]
                            envData.env_vars.BUILD_MVN_OPTS = buildMvnOpts + (envData.env_vars.BUILD_MVN_OPTS ?: '')
```
make sure `env_vars` is existing

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,413 @@
...
+                        def buildMvnOpts = "-Dproductized ${getAdditionalDefaultBuildMavenOpts()} "
+
+                        branchConfig.environments.each { envName, envData ->
+                            envData.env_vars = envData.env_vars ?: [:]
+                            envData.env_vars.BUILD_MVN_OPTS = buildMvnOpts + (envData.env_vars.BUILD_MVN_OPTS ?: '')
+                        }
+
+                        if (!branchConfig.environments.default) {
+                            // create default env if not already existing
+                            branchConfig.environments.default.env_vars.BUILD_MVN_OPTS = buildMvnOpts
```
radtriste: same goes here
```suggestion
                            branchConfig.environments.default = [
                                env_vars : [
                                    BUILD_MVN_OPTS: buildMvnOpts
                                ]
                            ]
```

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,420 @@
...
+// 2. QUARKUS_VERSION
+// 3. 
+
+seedConfig = [:]
+mainBranchConfig = [:]
+repositoriesToBranch = []
+
+pipeline{
+    agent{
+        label 'kie-rhel8 && !built-in'
```
radtriste:
```suggestion
        label 'rhel8 && !built-in'
```

we don't need 16GB for yaml manipulation :D

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
```diff
@@ -0,0 +1,420 @@
...
+                        enabledEnvs.each { envName, envData ->
+                            // ensure env is enabled
+                            envData.enabled = true
+
+                            envData.env_vars = envData.env_vars ?: [:]
+                            envData.env_vars.BUILD_MVN_OPTS = buildMvnOpts + (envData.env_vars.BUILD_MVN_OPTS ?: '')
+                        }
+
+                        // create default env if not already existing
+                        enabledEnvs.default = enabledEnvs.default ?: [:]
```
radtriste: why not doing this one just before the `// apply changes on remaining envs` part ?
that way, you don't need the next 2 lines :)

radtriste (approved):
2 minor comments

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
lampajr: fixed :smile:

dsl/seed/jenkinsfiles/Jenkinsfile.prod.prepare
lampajr: yeah that's true, fixed

radtriste (approved):

rgdoliveira (approved):

PR #988, opened 2023-06-13
KOGITO-7426 Fix pnpm/junit issue

radtriste:
JUnit was having heap space issues, most probably due to `node_modules` structure and number of files ...

This PR just remove the `node_modules` after the build to make sure the issue does not happen with Junit anymore

This also solves https://issues.redhat.com/browse/KOGITO-7926 by including the cypress junit.xml

### DIFF
--- a/.ci/jenkins/Jenkinsfile.buildchain
+++ b/.ci/jenkins/Jenkinsfile.buildchain
@@ -131,7 +131,10 @@ pipeline {
             post {
                 always {
                     script {
-                        sh 'find . -type d -name node_modules -exec rm -rf {} \\; || true' // Remove `node_modules` to avoid heap space issues with junit command thereafter
+                        // Remove `node_modules` to avoid heap space issues with junit command thereafter
+                        // Related to https://github.com/jenkinsci/junit-plugin/issues/478 and https://github.com/jenkinsci/junit-plugin/issues/467
+                        sh 'find . -type d -name node_modules -exec rm -rf {} \\; || true'
+
                         junit(testResults: '**/junit.xml, **/target/surefire-reports/**/*.xml, **/target/failsafe-reports/**/*.xml, **/target/invoker-reports/**/*.xml', allowEmptyResults: true)
                         archiveArtifacts(artifacts: '**/cypress/screenshots/**,**/cypress/videos/**', fingerprint: false, allowEmptyArchive: true)
                     }
--- a/.ci/jenkins/Jenkinsfile.buildchain
+++ b/.ci/jenkins/Jenkinsfile.buildchain
@@ -131,11 +131,9 @@ pipeline {
             post {
                 always {
                     script {
-                        dir(getProjectFolder()) {
-                            sh 'find . -type d -name node_modules -exec rm -rf {} \\; || true' // Remove `node_modules` to avoid heap space issues with junit command thereafter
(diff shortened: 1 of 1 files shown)


.ci/jenkins/Jenkinsfile.buildchain
```diff
@@ -130,8 +130,13 @@ pipeline {
...
                 always {
-                    junit(testResults: '**/target/surefire-reports/**/*.xml, **/target/failsafe-reports/**/*.xml, **/target/invoker-reports/**/*.xml', allowEmptyResults: true)
-                    archiveArtifacts(artifacts: '**/cypress/screenshots/**,**/cypress/videos/**', fingerprint: false, allowEmptyArchive: true)
+                    script {
+                        dir(getProjectFolder()) {
+                            sh 'find . -type d -name node_modules -exec rm -rf {} \\; || true' // Remove `node_modules` to avoid heap space issues with junit command thereafter
+                            junit(testResults: '**/junit.xml, **/target/surefire-reports/**/*.xml, **/target/failsafe-reports/**/*.xml, **/target/invoker-reports/**/*.xml', allowEmptyResults: true)
+                            archiveArtifacts(artifacts: '**/cypress/screenshots/**,**/cypress/videos/**', fingerprint: false, allowEmptyArchive: true)
+                        }
+                    }
```
lampajr: @radtriste With this change we will run junit/archive only for a single project (the one returned by `getProjectFolder`), which is different wrt previous approach that was archiving all matching artifacts on all projects, is this new behavior the one expected?

mareknovotny: i guess we need to exclude node_modules from junit, do not remove it as it can be long time operation to delay build job.

radtriste: you are right ... will set back full folder

radtriste: @mareknovotny if you managed to get it working, go ahead, I already tried a bit with no luck
honestly, this is a workaround that will make it for now and is not so long to delete ...

mareknovotny: the approach could be at https://github.com/jenkinsci/junit-plugin/issues/478 and https://github.com/jenkinsci/junit-plugin/issues/467 describes the OOO similar problem

What you tried as negation is not supported i believe. Anyway maybe the patterns need to be more specific without negation?

mareknovotny: I would say that removal should be just temporary if there is no other way ;) @radtriste

radtriste: > the approach could be at [jenkinsci/junit-plugin#478](https://github.com/jenkinsci/junit-plugin/issues/478) and [jenkinsci/junit-plugin#467](https://github.com/jenkinsci/junit-plugin/issues/467) describes the OOO similar problem
> 
> What you tried as negation is not supported i believe. Anyway maybe the patterns need to be more specific without negation?

more specific is difficult as the file is meant to be used by all projects to build easily in Jenkins
For now the removal seems a fair workaround. Gonna add the GH issue you mention in the comments

mareknovotny (approved):

lampajr (approved):

Sara4994 (approved):

PR #999, opened 2023-06-16
[BXMSPROD-2064] set untrusted input as intermediate env variable

lampajr:
**JIRA:** https://issues.redhat.com/browse/BXMSPROD-2064

**referenced Pull Requests:**
* https://github.com/kiegroup/kie-ci/pull/1514
* https://github.com/kiegroup/kogito-pipelines/pull/999
* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/2321

As suggested here: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable setting untrusted input as env variable

> NOTE: even if we are planning to remove composite actions from this repo I would like to apply the change also here even if it will be needed for short time

### DIFF
--- a/.ci/actions/parse-labels/action.yml
+++ b/.ci/actions/parse-labels/action.yml
@@ -36,5 +36,8 @@ runs:
         echo "targets=$(echo $targets)" >> $GITHUB_OUTPUT
     
     - name: Printing extracted targets
+      env:
+        # set untrusted input to intermediate env variable
+        TARGETS: ${{ steps.extract-targets.outputs.targets }}
       shell: bash
-      run: echo "Extracted target branches ${{ steps.extract-targets.outputs.targets }}"
+      run: echo "Extracted target branches $TARGETS"


radtriste (approved):

radtriste:
So putting them as env variable will remove the fact that it could be executed code ?

lampajr:
> So putting them as env variable will remove the fact that it could be executed code ?

yeah exactly, this is the first suggestion from GitHub doc. 
I already make a test locally and it worked as expected.

.ci/actions/backporting/action.yml
```diff
@@ -100,13 +100,19 @@ runs:
...
+        BODY: ${{ inputs.body }}
+        ORIGINAL_PR_URL: ${{ inputs.original-pr-url }}
       uses: carloscastrojumo/github-cherry-pick-action@v1.0.9
       with:
-        branch: "${{ inputs.target-branch }}"
-        title: "[${{ inputs.target-branch }}] ${{ inputs.title }}"
-        body: "**Backport:** ${{ inputs.original-pr-url }}\r\n\r\n${{ inputs.body }}"
+        branch: "$TARGET_BRANCH"
+        title: "[$TARGET_BRANCH] $TITLE"
+        body: "**Backport:** $ORIGINAL_PR_URL\r\n\r\n$BODY"
```
shubhbapna: According to the GitHub docs passing the untrusted input directly to the `with` portion of the action is safe:

> This approach is not vulnerable to the injection attack, as the context value is not used to generate a shell script, but is instead passed to the action as an argument

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-action-instead-of-an-inline-script-recommended

So IMO I don't think we have to add the `env` section here and we can keep it how it was before.

shubhbapna (changes requested):

.ci/actions/backporting/action.yml
lampajr: I agree with you @shubhbapna., TBH I added them also here more for readability rather than for need but yeah I can easily revert these changes as well

lampajr: reverted these changes :)

shubhbapna (approved):

PR #1082, opened 2023-09-12
PR multibranchPipelineJob

jstastny-cz:
Adding a template for multibranchPipelineJob which using the github branch source plugin DSL defines a job that periodically polls given repository for new PRs from origin and from forks and executes given pipeline script (from that particular repo).

### DIFF
--- a/dsl/seed/src/main/groovy/org/kie/jenkins/jobdsl/KogitoJobTemplate.groovy
+++ b/dsl/seed/src/main/groovy/org/kie/jenkins/jobdsl/KogitoJobTemplate.groovy
@@ -466,4 +466,41 @@ class KogitoJobTemplate {
         return "(.*${RegexUtils.getRegexFirstLetterCase('jenkins')},?.*(rerun|run) ${idStr}${testType}.*)"
     }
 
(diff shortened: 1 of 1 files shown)


dsl/seed/src/main/groovy/org/kie/jenkins/jobdsl/KogitoJobTemplate.groovy
```diff
@@ -466,4 +466,41 @@ class KogitoJobTemplate {
...
+        script.folder("pullrequest_jobs")
+        return script.multibranchPipelineJob("pullrequest_jobs/${Utils.getRepoName(script)}-pr")?.with {
+            factory {
+                workflowBranchProjectFactory {
+                    scriptPath("${jenkinsFilePath}")
+                }
+            }
+            branchSources {
+                github {
+                    id("${Utils.getRepoName(script)}") // IMPORTANT: use a constant and unique identifier
```
radtriste: No need to use gstring there and for the others below, you can directly use the result from the Utils method
```suggestion
                    id(Utils.getRepoName(script)}) // IMPORTANT: use a constant and unique identifier
```

radtriste (approved):

radtriste (approved):

PR #1091, opened 2023-09-20
kie-issues#578 install freetype package using dnf

jstastny-cz:
Adding freetype package which provides `libfreetype.so.6` that was reported as missing by Drools tests:
```
java.lang.UnsatisfiedLinkError: /home/nonrootuser/.sdkman/candidates/java/11.0.20-tem/lib/libfontmanager.so: libfreetype.so.6: cannot open shared object file: No such file or directory
```


baldimir (approved):

jstastny-cz:
Solving kiegroup/kie-issues#578

PR #1104, opened 2023-10-04
Remove maven actions to follow Apache policies

ricardozanini:
In this PR:

- Remove setup maven and maven settings XML actions to include custom bash scripts to do the same thing
- Review actions to cache maven download and fetch this repo for a settings.xml reference file and a script to install maven

TODO:
- [x] - Test on my fork

### DIFF
--- a/.ci/actions/maven/action.yml
+++ b/.ci/actions/maven/action.yml
@@ -30,10 +30,27 @@ runs:
         distribution: 'temurin'
         java-version: ${{ inputs.java-version }}
         check-latest: true
-    - name: Set up Maven
-      uses: stCarolas/setup-maven@v4.2
+    # We have to download the required scripts to our local workspace
+    - name: Checkout Scripts
+      uses: actions/checkout@v4
       with:
-        maven-version: ${{ inputs.maven-version }}
+        repository: apache/incubator-kie-kogito-pipelines
+        sparse-checkout: |
+          .ci/actions/maven
+        path: maven-tools
+    - name: Cache Maven Download
+      id: cache-maven-download
+      uses: actions/cache@v3
+      with:
+        key: ${{ runner.os }}-maven-${{ inputs.maven-version }}
+        path: |
+          ~/.maven/download/
+    - name: Set up Maven
+      shell: bash
+      run: |
+        ./maven-tools/.ci/actions/maven/install-maven.sh "$HOME/.local/mvn"
+        echo "$HOME/.local/mvn/bin" >> $GITHUB_PATH
+        
     - name: Cache Maven packages
       uses: actions/cache@v3
       with:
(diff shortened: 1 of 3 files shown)


jstastny-cz:
We're adapting for Apache projects, is there expectation to support also non transfered kiegroup projects? If so, we should somehow parameterize repository information. 
If it's only for Apache projects, then we could replace the URLs.

Otherwise looks good, installation of maven could be done using sdkman :-) (if you have any reservations to its use in CI I'd like to learn.)

I don't have any insight into before/after comparison of the configuration done using these actions, but seems legit.

ricardozanini:
It's working now, you can see it here: https://github.com/ricardozanini/java-helloworld/actions

@jstastny-cz thanks! Replies inline.

> We're adapting for Apache projects, is there expectation to support also non transfered kiegroup projects? If so, we should somehow parameterize repository information.

Not sure what you mean. I don't think we need to parameterize repository information, can you give more information? Where, exactly?

> installation of maven could be done using sdkman

I don't have any restrictions, we can do it using whatever, including the GHA Java Action, but since we need a specific Maven version ('cause related/dependency projects), Tristan used a third-party action. If we can specify a version and have it added to classpath, feel free to add a commit to this PR using sdkman.

I just would like to move forward fast to avoid being blocked by CI. :/

jstastny-cz:
All good, for the repositories I meant in the Maven settings template - I saw there jboss nexus URLs. This should be changed to Apache nexus URLs or parameterized by the actions config?

ricardozanini:
> All good, for the repositories I meant in the Maven settings template - I saw there jboss nexus URLs. This should be changed to Apache nexus URLs or parameterized by the actions config?

OH SURE

We can update the template in a later PR adding Apache URLs! :D

jstastny-cz:
I would suggest replacing as part of this one to prevent possible use of stale snapshots. There is no need to have jboss URL there already.

ricardozanini:
> I would suggest replacing as part of this one to prevent possible use of stale snapshots. There is no need to have jboss URL there already.

Cool! Can you give me the updated URLs or just add a suggestion to change here? I understand that this is a super simple change to do, right? Thank u!

.ci/actions/maven/action.yml
```diff
@@ -48,70 +65,11 @@ runs:
...
-              }
-            }
-          ]
-        plugin_groups: >
-          [
-            "org.zanata"
-          ]
+      shell: bash
+      run: |
+        cp maven-tools/.ci/actions/maven/settings-template.xml ${HOME}/.m2/settings.xml
```
baldimir, line 70: I think the custom settings.xml generation can be removed completely. Each repository has the configuration of the repositories in their root pom, e.g. here it is in kogito-runtimes https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/pom.xml#L75. I think that way the deployment etc. can be controlled more easily, than in a custom action like this. What do you think please?

baldimir (approved):

jstastny-cz (approved):
LGTM. Let's follow the PR executions once this is in place and closely investigate any differences in PR checks outcome GHA vs Jenkins.

PR #1107, opened 2023-10-06
Strict Build Chain version to whitelisted

ricardozanini:
In this PR, we strict the version used in [Build Chain](https://github.com/kiegroup/github-action-build-chain) to the one whitelisted by Apache.

### DIFF
--- a/.ci/actions/build-chain/action.yml
+++ b/.ci/actions/build-chain/action.yml
@@ -33,7 +33,7 @@ runs:
   steps:
     - name: Build Chain Tool Execution
       id: build-chain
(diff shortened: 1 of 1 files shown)


radtriste:
Please do not change that. It should get already automatically the latest buildchain 3 version
But it takes some time for the npm release to propagate

jstastny-cz:
Well, consider we need to white-list explicit version in apache organization, check https://issues.apache.org/jira/browse/INFRA-25050

ricardozanini:
@jstastny-cz let's wait for the 3.x version to be whitelisted, then we can close this PR without merging it.

PR #1112, opened 2023-10-25
kie-issues#651: introduce DISABLE_PR_CHECK config env variable

jstastny-cz:
apache/incubator-kie-issues#651

Introducing a utility method to disable PR checks for given project.
It disables them by default in "non-production" environment, when the seed or seed config author is not apache org to prevent unintended triggering of PR checks from test/temporary dsl config.


jstastny-cz:
Validated in a test pipeline generation, turning to regular PR.

baldimir (approved):

PR #1125, opened 2023-11-23
kie-issues#710: turn kogito-ci-build to ubuntu dind image

jstastny-cz:
apache/incubator-kie-issues#710

Switching the kogito-ci-build into ubuntu based dind image which will isolate container docker environment from the host one (as opposed to our previous solution by passing docker.sock).

The pipelines are currently generated with a fixed kogito-ci-build image tag, thus by merging this, we'll allow the build pipeline to build and push the new image under main-latest tag, only when that's successful, we'll switch pipelines back to using main-latest again. The reason is that docker args will change together with the image.

### DIFF
--- a/.ci/jenkins/Jenkinsfile.build-kogito-ci-image
+++ b/.ci/jenkins/Jenkinsfile.build-kogito-ci-image
@@ -35,7 +35,7 @@ pipeline {
                         checkout scm
 
                         sh """
-                            docker build --build-arg SDKMAN_JAVA=11.0.20-tem -t ${env.IMAGE_NAME_TAG} -f apache-nodes/Dockerfile.kogito-ci-build .
+                            docker build --build-arg SDKMAN_JAVA=11.0.20-tem -t ${env.IMAGE_NAME_TAG} -f apache-nodes/Dockerfile.kogito-ci-build ./apache-nodes
                             docker tag ${env.IMAGE_NAME_TAG} ${env.IMAGE_NAME}:${GIT_COMMIT}
                             docker tag ${env.IMAGE_NAME_TAG} ${env.IMAGE_NAME}:${BRANCH_NAME}-latest
                         """
@@ -61,7 +61,7 @@ pipeline {
             agent {
                 docker {
                     image env.IMAGE_NAME_TAG
-                    args '-v /var/run/docker.sock:/var/run/docker.sock --group-add docker --group-add input --group-add render'
+                    args '--privileged --group-add docker'
                 }
             }
             steps {
--- a/.ci/jenkins/Jenkinsfile.build-kogito-ci-image
+++ b/.ci/jenkins/Jenkinsfile.build-kogito-ci-image
@@ -35,7 +35,7 @@ pipeline {
                         checkout scm
 
                         sh """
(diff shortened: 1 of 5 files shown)


jstastny-cz:
The image has been built as part of https://ci-builds.apache.org/job/KIE/job/kogito/job/main/job/branch_jobs/job/kogito-ci-build-image/job/kie-issues_710/2 and is available for testing at quay.io/kiegroup/kogito-ci-build:kie-issues_710-latest

baldimir (approved):

apache-nodes/Dockerfile.kogito-ci-build
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi:9.2
+FROM cruizba/ubuntu-dind:latest
```
radtriste, line 1: Is that ok using an external no "official" image ?

dsl/scripts/pr_check.groovy
```diff
@@ -31,7 +31,7 @@ dockerArgs = [
 ] + dockerGroups.collect { group -> "--group-add ${group}" }
 
 void launch() {
-    String builderImage = 'quay.io/kiegroup/kogito-ci-build:main-latest'
+    String builderImage = 'quay.io/kiegroup/kogito-ci-build:19a0b303bc64f473a01f5fa5bacde822f10b4946' // last main-latest based on ubi
```
radtriste, line 34: Is this temporary ?

jstastny-cz: Yes temporary, it was a leftover from previous change of pipeline configuration.

Once this new image is pushed as main-latest, this will be adjusted accordingly including the changed args.

apache-nodes/Dockerfile.kogito-ci-build
jstastny-cz: Seems to be well-maintained. https://hub.docker.com/r/cruizba/ubuntu-dind/tags

When it comes to security, I've reviewed the current codebase, all the changes seem valid to me, though I am no expert on container security. https://github.com/cruizba/ubuntu-dind/blob/master/ubuntu-jammy.Dockerfile

One thing we could do is to specify an exact version of the base image, so that we have control over what we build on? With understanding that when upgrading, we should review the image we're building on.

radtriste: Using a fix version would miss the bug fixes. Sadly there is no distribution image. This could be suggested to the repository owner.
Code seems legit indeed. But as you are giving privileged rights to the container, we should be careful

jstastny-cz: I am asking for sysbox to be installed on the hosts, let's see.

.ci/jenkins/Jenkinsfile.build-kogito-ci-image
```diff
@@ -61,7 +61,7 @@ pipeline {
             agent {
                 docker {
                     image env.IMAGE_NAME_TAG
-                    args '-v /var/run/docker.sock:/var/run/docker.sock --group-add docker --group-add input --group-add render'
+                    args '--privileged --group-add docker'
```
rodrigonull, line 64: If this image is going to also be used for running tests using selenium / chrome I suggest to add this arg: `--shm-size=2gb`

rodrigonull (approved):

PR #1140, opened 2023-12-08
kie-issues#747: tweak CI image docker startup

jstastny-cz:
Fixes
* apache/incubator-kie-issues#747

* Adding an extra script to kogito-ci-build image to allow waiting for dockerd to be started.
Reason is that Jenkins does not wait for entrypoint.sh to finish before it starts executing steps towards it. Thus sometimes the docker is not yet started and docker socket available.
* Adding also pipeline util method util.waitForDocker() which first sleeps for short period of time before invoking the wait script against the container - in some cases the waiting script's invocation seemed to be too early to be successful - jenkins would just hang.
Thanks to library definition changes in the Jenkinsfile.build-kogito-ci-image the jenkins library change was properly tested.
* Also included change to change default shell to bash, instead of dash.

### DIFF
--- a/.ci/jenkins/Jenkinsfile.build-kogito-ci-image
+++ b/.ci/jenkins/Jenkinsfile.build-kogito-ci-image
@@ -1,12 +1,12 @@
 import org.jenkinsci.plugins.workflow.libs.Library
 
-@Library('jenkins-pipeline-shared-libraries')_
-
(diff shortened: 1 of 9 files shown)


jstastny-cz:
Disregard the jenkins build, it's mistakenly triggered from a testing location.

PR #1143, opened 2023-12-12
Remove set -e from scripts sourced from .bashrc

### LINKED ISSUE #756: remove set -e from scripts sourced from .bashrc in kogito-ci-build image
Currently, kogito-ci-build image sources 
```
RUN echo "source wait-for-docker.sh" >> ~/.bashrc
```
a script which has `set -e` in it.

This results in the interactive shell session being itnerrupted on a non-zero return code.

E.g. wrongly shaped call to `git config` is a reproducer.

Should not be affecting CI as the shell session there is not interactive.

rodrigonull:
CLOSES https://github.com/apache/incubator-kie-issues/issues/756

### DIFF
--- a/apache-nodes/entrypoint.sh
+++ b/apache-nodes/entrypoint.sh
@@ -1,6 +1,4 @@
 #!/bin/bash
-set -e
-
 # cgroup v2: enable nesting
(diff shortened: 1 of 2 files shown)


jstastny-cz (approved):
LGTM. Thank you.

PR #1179, opened 2024-02-07
Fix #1178 - Remove Kogito Operator pipelines and references

### LINKED ISSUE #1178: Remove Kogito Operator Pipelines
See https://github.com/apache/incubator-kie-kogito-operator/issues/1537

ricardozanini:
See https://github.com/apache/incubator-kie-kogito-operator/issues/1537

I've kept the node and the image that builds the operator 'cause we reuse it for the SonataFlow operator too.

Also, the example-images pipeline CAN be disabled since we only use it for the Kogito Operator BDD. But we need a new assessment on it to understand if they are being used somewhere else. We can open another PR to take care of it.


jstastny-cz:
@ricardozanini IMHO the example images were exclusively for operator BDD tests. Let's ask @radtriste to confirm and we could get rid of it already.

kogito-jenkins-node/image.yaml
```diff
@@ -1,7 +1,7 @@
 name: quay.io/kiegroup/kogito-jenkins-node
 version: "1.40.0"
 from: quay.io/openshift/origin-jenkins-agent-base:4.14
-description: Image used by dynamic jenkins agent to build kogito-cloud-operator using podman
+description: Image used by dynamic jenkins agent to build Kogito Operators using podman
```
jstastny-cz: I don't think we use this image in apache CI at all ....can you double-check? Or is that used elsewhere?

radtriste:
As far as I remember, they were mostly used for the bdd tests. We were indeed pushing them but I don't think that we were mentioning them into the documentation and/or examples' readmes.
That should be easy to check (look for 'quay.io/kiegroup/example-*' string

radtriste (approved):

kogito-jenkins-node/image.yaml
radtriste: I think the `kogito-jenkins-node` can be deleted. It was used in IBM Jenkins for external access. But now we are using Apache Jenkins

rodrigonull (approved):
LGTM.

ricardozanini:
@radtriste @jstastny-cz many thanks for the updates! I'll remove the Jenkins Node and the BDD tests. I'll also squash this commit to make it easier to grab anything back if needed.

ricardozanini:
@jstastny-cz @radtriste can you please take a look to make sure that my exclusions are alright? Thank you!

radtriste (approved):

jstastny-cz (approved):

PR #1193, opened 2024-05-07
kie-issues#872: adjust branch seed

jstastny-cz:
Follows-up on 
* apache/incubator-kie-issues#872
* apache/incubator-kie-drools#5863

This change relaxes relation between `buildchain-config` and `repositories` sections of branch.yaml config:
* buildchain-config repository does not need to be in repositories section


rodrigonull (approved):
LGTM. Thanks!

baldimir (approved):

PR #1196, opened 2024-05-15
kie-issues#1212: fix clean* pipelines from tools directory

jstastny-cz:
Fixes:
* apache/incubator-kie-issues#1212

Add missing Library declaration and adjust docker config in tools/clean* pipelines.


cimbalek (approved):
LGTM

PR #1210, opened 2024-06-20
NO-ISSUE: fix buildchain for native build

jstastny-cz:
Fixing buildchain jenkinsfile for native build (when snapshot deploy is disabled).

Problem was that in case of build-and-test jobs variants we're not passing the credentials via env at all, thus the withCredentials block need to be conditional on whether it is deploy or not.


rodrigonull (approved):
LGTM

PR #1218, opened 2024-07-01
Disable kogito-serverless-operator pipelines

rodrigonull:
Kogito serverless operator code has been moved to KIE tools for the Apache 10 release.

We need to disable the CI pipelines for this project to avoid having images published from these jobs until we have the Apache 10 release and move the code back to the original git repository.


jstastny-cz (approved):
LGTM.

thiagoelg (approved):

PR #1233, opened 2024-08-01
Fix for kie issue #1403

pibizza:
There is a missing docker directive inside the agent directive. This should fix it.


jstastny-cz (approved):
LGTM.

rodrigonull (approved):

PR #1234, opened 2024-08-01
Fix for kie issue #1405. Removing quay image cleanup since we are no longer using quay.

pibizza:
This removes the job to cleanup the images on quay. There will be also a backport for 10.x branch.


jstastny-cz (approved):
LGTM. Thank you.

rodrigonull (approved):
LGTM

PR #1249, opened 2024-09-17
setting compare to non failure

elguardian:


elguardian:
@porcelli @martinweiler tyring to unstack work without compromising pr. This allows to know exactly which modules are failing and do in parallel thoses fixes if require. Once the module are fixed we can go on and set this to failure = true

jstastny-cz (approved):
LGTM.

gitgabrio (changes requested):
Please leave the exclusions as they are.
We already know those modules are not reproducible "by-design" - no need to waste time building them twice

gitgabrio (approved):
Thx @elguardian

PR #1270, opened 2024-12-05
[10.0.x] [incubator-kie-issues-1618] Fix NOTICE year with range

tkobayas:
- https://github.com/apache/incubator-kie-issues/issues/1618

See https://github.com/apache/incubator-kie-issues/issues/1618#issuecomment-2516665249

I didn't file a PR for main, because NOTICE in main looks old, so I assume it will be overwritten with NOTICE in 10.0.x

### DIFF
--- a/tools/zip-sources-files/NOTICE
+++ b/tools/zip-sources-files/NOTICE
@@ -1,5 +1,5 @@
 Apache KIE
-Copyright 2024 The Apache Software Foundation
+Copyright 2023-2024 The Apache Software Foundation
 
 This product includes software developed at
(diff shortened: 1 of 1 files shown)


jomarko (approved):

yesamer (approved):

tkobayas:
Hi @porcelli , I'm merging this PR.

Rc4 NOTICE still has `2024`. but I believe it's not a blocker, because "Malformed NOTICE" is minor issue : https://issues.apache.org/jira/browse/LEGAL-469

PR #1282, opened 2025-08-20
Build-chain kogito-apps - Add after script for reproducible build check

jstastny-cz:
Adjusting kogito-apps build-chain build scripts to kogito-runtimes ones regarding the reproducible build checks.


elguardian (approved):
LGTM @jstastny-cz great work !

PR #1296, opened 2026-02-24
NO-ISSUE: workaround buildchain jenkinsfile reference to GITHUB_WORKSPACE

jstastny-cz:
In attempt to reduce size of logs, the applicable build-chain config was adjusted to write mvn dependency:tree verbose output into a file inside workspace. The primary focus was GitHub PR checks, so the destination is now defined as env.GITHUB_WORKSPACE.

While this fits GitHub actions, obviously there's problem in jenkins, where the same build-chain configuration file is shared by the nightly jobs.
There were two options:
1. refactor the build-chain configuration to use a different env variable, which then would need to be explicitly set in all the workflows (quite many all around the repos).
2. adjust Jenkinsfile that uses the shared build-chain configuration to "fake" env.GITHUB_WORKSPACE while providing explanation in a comment. Which isolates the change to a single Jenkinsfile.

The PR here implements option 2 from the above.

### DIFF
--- a/dsl/seed/jenkinsfiles/Jenkinsfile.buildchain
+++ b/dsl/seed/jenkinsfiles/Jenkinsfile.buildchain
@@ -57,6 +57,10 @@ pipeline {
         timestamps()
         timeout(time: timeoutValue, unit: 'MINUTES')
     }
+    environment {
+        GITHUB_WORKSPACE = "${WORKSPACE}" // faking GITHUB_WORKSPACE env var for the build-chain config mvn dependency:tree output
+        DEPENDENCY_TREE_FILE = "mvn_dependency_tree.txt"
+    }
     stages {
         stage('Initialize') {
             steps {


dsl/seed/jenkinsfiles/Jenkinsfile.buildchain
```diff
@@ -57,6 +57,10 @@ pipeline {
         timestamps()
         timeout(time: timeoutValue, unit: 'MINUTES')
     }
+    environment {
+        GITHUB_WORKSPACE = "${WORKSPACE}" // faking GITHUB_WORKSPACE env var for the build-chain config mvn dependency:tree output
+        DEPENDENCY_TREE_FILE = "mvn_dependency_tree.txt"
```
gitgabrio, line 62: HI @jstastny-cz 
Many thanks for the quick fix.
My only doubt is that, in my original PRs, I had to create different "names" to avoid clashing:
is there a way to check that for this PR ?

jstastny-cz: You also used `-DappendOutput=true` , so it should not be problem during individual configuration's execution. And even if there's a matrix job in Jenkins (which I don't think there is such), each workspace is isolated in Jenkins. Perhaps not in GitHub action and that's why you introduced the distinction?

gitgabrio: Yup, exactly.

gitgabrio (approved):
Thanks @jstastny-cz !

domhanak (approved):

PR #1301, opened 2026-05-13
[incubator-kie-issues#2029] Attempt remove full profile building of kogito-runtimes as upstream module

gitgabrio:
Removing "full" profile when building kogito-runtimes upstream from kogito-apps.

It was probably needed to build serverless-workflow-python modules, but those are not there anymore, and the mix of quickly and full profile may create unexpected issues


yesamer (approved):

jstastny-cz (approved):
Makes sense, thank you @gitgabrio .

